PR #89, opened 2023-08-30
Implement RandomPolicy

Junyoungpark:
## Description

This update contains the implementation of the `RandomPolicy` class
The random policy constructs solutions with randomly chosen actions.

## Motivation and Context

Having a random policy can be useful during the development of a new environment to perform the sanity check.

## Types of changes

- [X] New feature (non-breaking change which adds core functionality)

## Checklist

- [X] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [X] I have updated the documentation accordingly.

### DIFF
--- a/rl4co/models/nn/utils.py
+++ b/rl4co/models/nn/utils.py
@@ -58,15 +58,19 @@ def random_policy(td):
     return td
 
 
-def rollout(env, td, policy):
+def rollout(env, td, policy, max_steps: int = 10_000):
     """Helper function to rollout a policy. Currently, TorchRL does not allow to step
     over envs when done with `env.rollout()`. We need this because for environements that complete at different steps.
     """
     actions = []
+    steps = 0
     while not td["done"].all():
         td = policy(td)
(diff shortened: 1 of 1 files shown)


rl4co/models/nn/utils.py
```diff
@@ -55,17 +58,50 @@ def random_policy(td):
...
+    """
+
+    def __init__(self, env_name=None):
+        super().__init__()
+        self.env_name = env_name
+
+    def forward(
+        self,
+        td: TensorDict,
+        env: Union[str, RL4COEnvBase] = None,
```
fedebotu, line 98: How about setting the`max_steps` to `inf` by default?
Since there may be cases in which one wants to test a very large scale problem, but this `max_steps` is kinda hardcoded. Or, perhaps we can log a warning if this happens

Junyoungpark: I set the default value of `max_steps` as `None`. If `max_steps=None`, rollout decodes until it finds the environments are done.

PR #91, opened 2023-09-10
Pass tensordict as an additional input of _precompute_cache

Junyoungpark:
## Description

Pass td as an additional input of `_precompute_cache` to allow better  flexibility of `_precompute_cache` in the children classes

## Types of changes

- [X] New feature (non-breaking change which adds core functionality)

## Checklist

- [X] I have updated the tests accordingly (*required for a bug fix or a new feature*).

### DIFF
--- a/rl4co/models/zoo/common/autoregressive/decoder.py
+++ b/rl4co/models/zoo/common/autoregressive/decoder.py
@@ -3,7 +3,6 @@ from typing import Tuple, Union
 
 import torch
 import torch.nn as nn
-
 from einops import rearrange
 from tensordict import TensorDict
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


Junyoungpark:
@fedebotu would you take a look at this?
I found passing td as an extra argument for `_precompute_cache` can be useful to design a subclass of `AutoregressivePolicy` with the extended features.

fedebotu:
I left a minor comment as review - I think it may be useful to substitute the `PrecomputedCache` into something more TensorDict-like in the future - for instance, `TensorClass` may do

Junyoungpark:
I leave the reply comments in the code!
yeah, I agree with the suggestion!

PR #93, opened 2023-09-13
[Refactor] move A2C in separate folder for clarity

fedebotu:
## Description

Move Critic from REINFORCE to A2C

## Motivation and Context

It may be confusing for some to see "critic" as a baseline for REINFORCE when, indeed, it is equal to A2C. This is a way to make it easier for newcomers to grasp some concepts.

## Types of changes

- [x] Minor (refactoring)

## Checklist

- [ ] Testing.


fedebotu:
Not a high priority for the moment. We can keep this PR as a template in case we want to make A2C methods clearer in the future

PR #98, opened 2023-10-13
Fix order of shape for unbatchify in POMO

hyeok9855:
## Description

Fix order of argument `shape` for `unbatchify` function in POMO Model.

## Types of changes

The original argument for `shape` was `(n_start, n_aug)`, with which the shape of tensors will be changed to `[batch_size, n_start, n_aug]`. But we want it to be `[batch_size, n_aug, n_start]` so that it can mirror the batchify operation appropriately.

## Motivation and Context

Note that the bug will not harm your training procedure or maximum reward calculation, but it should be fixed because it is not what we wanted, and make debugging difficult.

## Checklist

**I found that there's no test code for POMO though :(**

### DIFF
--- a/rl4co/models/zoo/pomo/model.py
+++ b/rl4co/models/zoo/pomo/model.py
@@ -71,24 +71,24 @@ class POMO(REINFORCE):
         out = self.policy(td, self.env, phase=phase, num_starts=n_start)
 
         # Unbatchify reward to [batch_size, num_augment, num_starts].
-        reward = unbatchify(out["reward"], (n_start, n_aug))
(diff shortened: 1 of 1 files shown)


hyeok9855:
By the way, I guess the selecting best actions are meaningless because it is not used or returned.

fedebotu:
Thanks!
Yes as we discussed, this part was working anyways since it's only during inference (so it doesn't matter in which order how you take the max(max()). But it's better to fix ofc 👌
Btw could you have a look at SymNCO as well to see if they there is the same problem?

PR #99, opened 2023-10-18
[BugFix] fix reward dim problem of shared baseline

cbhua:
In the shared baseline function, the size of the `reward` now is `[batch_size]`.

### DIFF
--- a/rl4co/models/rl/reinforce/baselines.py
+++ b/rl4co/models/rl/reinforce/baselines.py
@@ -53,8 +53,8 @@ class NoBaseline(REINFORCEBaseline):
 class SharedBaseline(REINFORCEBaseline):
     """Shared baseline: return mean of reward as baseline"""
 
-    def eval(self, td, reward, env=None, on_dim=1):  # e.g. [batch, pomo, ...]
-        return reward.mean(dim=on_dim, keepdims=True), 0
+    def eval(self, td, reward, env=None):  # e.g. [batch, pomo, ...]
+        return reward.mean(), 0
 
 
 class ExponentialBaseline(REINFORCEBaseline):


fedebotu:
Let's have a closer look first, it might work but not for all baselines - if the baseline is multiple (i.e. symmetric and multistart) we should need the `dim` on which to operate the `mean` operation - for example [here](https://github.com/alstn12088/Sym-NCO/blob/02b1c98a072548829c15ffbf543dbe28ec8bdd4e/Sym-NCO-POMO/CVRP/CVRPTrainer.py#L220)  
Also @hyeok9855 you might want to have a look at this since you are working with POMO!

rl4co/models/rl/reinforce/baselines.py
```diff
@@ -53,8 +53,8 @@ def eval(self, td, reward, env=None):
 class SharedBaseline(REINFORCEBaseline):
     """Shared baseline: return mean of reward as baseline"""
 
-    def eval(self, td, reward, env=None, on_dim=1):  # e.g. [batch, pomo, ...]
-        return reward.mean(dim=on_dim, keepdims=True), 0
+    def eval(self, td, reward, env=None):  # e.g. [batch, pomo, ...]
+        return reward.mean(), 0
```
hyeok9855, line 57: Isn't the code after the change wrong?

image

fedebotu: Yes, that's what I think. @cbhua wants to fix some edge case that causes a bug, but it looks like the problem may not be the shared baseline itself

fedebotu: @cbhua any news on this PR? Since the above seems to be correct, we may close this PR

PR #110, opened 2024-01-23
[Feature] Add Beam Search as decoding strategy

### LINKED ISSUE #109: [Feature Request] Add Beam Search as decoding strategy
Hey there, 

first, big thanks for making this awesome library open source. I wish I would have discovered it earlier, would have probably saved me several hours of work ;) 

One thing I was missing though is the beam search decoding strategy. Do you already plan on adding it in the future? Otherwise I would be happy to help with that 

## Motivation
Beam search can lead to significant improvements of the solution quality of the trained model (e.g. Kool et al., 2018)

## Solution

This would probably require a base decoding strategy class, where the different strategies (greedy, sampling, beam search) would define a *step* function, called in every iteration of the autoregressive decoder. Also, something like a *post_decoder_hook*, where the backtracking of the beam search (and the concatenation of the actions in the other sampling strategies) takes place, would be required. 

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

LTluttmann:
## Description

- added base DecodingStrategy class which defines the following functions
     1. __pre_decoder_hook__: called before the while loop in the autoregressive decoder
     2. __step__: is called in every iteration of the while loop and samples an action given the log probabilities. The actual logic is specified by the subclasses through a *_step* function
     3. __post_decoder_hook__: called right after the while loop, mainly used to concatenate outputs

- implemented the greedy and sampling decoding strategies using the DecodingStrategy baseclass. They also support multistart by simply specifying "multistart_greedy" as decode_type for example. From a user perspective, nothing changes here.

- added beam search using the DecodingStrategy baseclass overwriting the following functions:
     1. __pre_decoder_hook__: similar to the multistart options, the beam width (if not specified) is determined and the actions of the first iteration are simply the first *beam_width* nodes (this could be improved in the future)
     2. **_step**: performs the actual beam search
     3. __post_decoder_hook__: performs back tracking and brings the solution in the right format

- added new test function to check that the new approach doesn't break anything

## Motivation and Context

close #109 

- [x] I have raised an issue to propose this change

## Types of changes

## Checklist

### DIFF
--- a/rl4co/models/zoo/common/autoregressive/decoder.py
+++ b/rl4co/models/zoo/common/autoregressive/decoder.py
@@ -10,12 +10,12 @@ from torch import Tensor
 
 from rl4co.envs import RL4COEnvBase, get_env
 from rl4co.models.nn.attention import LogitAttention
+from rl4co.models.nn.dec_strategies import DecodingStrategy, get_decoding_strategy
 from rl4co.models.nn.env_embeddings import env_context_embedding, env_dynamic_embedding
 from rl4co.models.nn.env_embeddings.dynamic import StaticEmbedding
-from rl4co.models.nn.utils import decode_probs, get_log_likelihood
-from rl4co.utils.ops import batchify, get_num_starts, select_start_nodes, unbatchify
+from rl4co.models.nn.utils import get_log_likelihood
+from rl4co.utils.ops import batchify, select_start_nodes, unbatchify
 from rl4co.utils.pylogger import get_pylogger
-from rl4co.models.nn.dec_strategies import DecodingStrategy, get_decoding_strategy
 
 log = get_pylogger(__name__)
 
@@ -146,7 +146,6 @@ class AutoregressiveDecoder(nn.Module):
             actions: Tensor of shape (batch_size, seq_len) containing the sampled actions
             td: TensorDict containing the environment state after decoding
         """
-
         # Instantiate environment if needed
         if isinstance(env, str):
             env_name = self.env_name if env is None else env
@@ -157,7 +156,9 @@ class AutoregressiveDecoder(nn.Module):
 
         # setup decoding strategy
         self.decode_strategy: DecodingStrategy = get_decoding_strategy(decode_type)
-        td, env, num_starts = self.decode_strategy.pre_decoder_hook(td, env, num_starts=num_starts)
+        td, env, num_starts = self.decode_strategy.pre_decoder_hook(
+            td, env, num_starts=num_starts
+        )
 
         # Main decoding: loop until all sequences are done
         while not td["done"].all():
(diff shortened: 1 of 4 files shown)


rl4co/models/zoo/common/autoregressive/decoder.py
```diff
@@ -133,74 +136,38 @@ def forward(
...
-            td.set("action", action)
-            td = env.step(td)["next"]
-            log_p = torch.zeros_like(
-                td["action_mask"], device=td.device
-            )  # first log_p is 0, so p = log_p.exp() = 1
-
-            outputs.append(log_p)
-            actions.append(action)
+        # setup decoding strategy
+        self.decode_strategy: DecodingStrategy = get_decoding_strategy(
```
fedebotu, line 158: I wonder if instantiating the strategy each time has some impact on speed? I guess not much. In case there is indeed some difference, it might be worth it to "cache" the strategy - i.e. since we are saving it in `self`, then we could do something like:

```python

if self.decode_strategy.name != decode_type:
  self.decode_strategy: DecodingStrategy = get_decoding_strategy(
              decode_type, **strategy_kwargs
          )
```

fedebotu (commented):
Great job on the PR! I left some mostly minor comments

fedebotu:
If you have some time, how about making a simple notebook (like [this](https://github.com/ai4co/rl4co/blob/main/notebooks/1-quickstart.ipynb)) with training and evaluation on different decoding strategies? I think it would make a great tutorial 😄

rl4co/models/zoo/common/autoregressive/decoder.py
LTluttmann: Thanks for the quick review @fedebotu! Great catch, making the DecodingStrategy a subclass of nn.Module is indeed not necessary. After removing that, I did some profiling to check the impact of instantiating the strategy each time. Results indicate a super minor impact on speed, please see the image below. For the test, I implemented an optional cache, similar to what you proposed, and checked the speed with and without caching the strategy. 
From a readability point-of-view, I would argue that the cache is not necessary. What do you suggest? :)

Bildschirmfoto 2024-01-29 um 13 02 20

fedebotu: Now that the class instantiation is simple, there is basically no overhead and much cleaner code. Let's keep it as is!

fedebotu:
Thanks a lot @LTluttmann for your contribution to RL4CO!
If you are interested in contributing with new features for the library, feel free to contact us (either here or through the [AI4CO community Slack](https://join.slack.com/t/rl4co/shared_invite/zt-1ytz2c1v4-0IkQ8NQH4TRXIX8PrRmDhQ)), we have a looong todo list ;)

PR #112, opened 2024-01-28
Adding JSSP environment

DavideTr8:
## Description

Added the JSSP env from Learning to Dispatch.

## Motivation and Context

We wanted to have a JSSP environment from a published paper.

- [ ] I have raised an issue to propose this change ([required](https://github.com/Daewoong-AI/DABA6-article-crawler/issues) for new features and bug fixes)

## Types of changes

## Checklist


fedebotu (commented):
Great job! 💪🏼

rl4co/envs/scheduling/jssp.py
```diff
@@ -0,0 +1,665 @@
...
+
+class Configs:
+    low = 1
+    high = 99
+    et_normalize_coef = 1000
+    rewardscale = 0.0
+    init_quality_flag = False
+
+
+configs = Configs()
```
fedebotu, line 19: About config: is this supposed not to be ever changed by the user?

Because if it can be changed, isn't it better to set these as hyperparameters to the `JSSPEnv` class?

rl4co/envs/scheduling/jssp.py
```diff
@@ -0,0 +1,665 @@
...
+    Note:
+        -
+    """
+
+    name = "jssp"
+
+    def __init__(self, num_jobs, num_machines, **kwargs):
+        super().__init__(**kwargs)
+        self.batch_size = torch.Size([1])
+        adjacency_spec = DiscreteTensorSpec(
```
fedebotu, line 39: [Minor] For specs, I think it would be cleaner to move them to a `_make_spec(...)` class as done in the [tutorial](https://pytorch.org/rl/tutorials/pendulum.html#:~:text=def%20_make_spec(self%2C%20td_params)%3A)

rl4co/envs/scheduling/jssp.py
```diff
@@ -0,0 +1,665 @@
...
+        self.durations_cp = torch.zeros(
+            (self.num_jobs, self.num_machines),
+            dtype=torch.float32,
+            device=self.device,
+        )
+
+    def done(self):
+        if len(self.partial_sol_sequence) == self.num_tasks:
+            return torch.tensor(True)
+        return torch.tensor(False)
```
fedebotu, line 127: Important note:
In general, whenever there is a `torch.tensor(...)`, `torch.ones(...)` etc to be instantiated , it is better to do so on the device. Since we use GPU for training and we can move the envs there for offloading the CPU, it would be best to instantiate directly there.
For instance:

```python
...
return torch.tensor(False, device=self.device)
```

The same applies to other tensors instantiated this way - at the moment, trying to set the env on GPU raises some device mismatch error, but it can be easily fixed :)

rl4co/envs/scheduling/jssp.py
```diff
@@ -0,0 +1,665 @@
...
+            shape=self.batch_size,
+            device=self.device,
+            dtype=torch.int64,
+        )
+
+        self.num_jobs = num_jobs
+        self.num_machines = num_machines
+        self.num_tasks = self.num_jobs * self.num_machines
+        # the task id for first column
+        self.first_col = torch.arange(
```
fedebotu, line 91: Here it seems that there are some tensors declared inside the  `__init__` signature; this makes the environment _stateful_ (as explained [here](https://pytorch.org/rl/tutorials/pendulum.html#:~:text=To%20give%20a,API%20in%20TorchRL.)) .

I don't hold a grudge against stateful envs, but I wonder whether it would be better (i.e. more efficient) to have them as in the [pendulum example](https://pytorch.org/rl/tutorials/pendulum.html), which is, simply passed upon generation?

rl4co/envs/scheduling/jssp.py
```diff
@@ -0,0 +1,665 @@
...
+            ),
+            dim=1,
+        )
+        reward = -(self.LBs.max() - self.max_end_time)
+        if reward == 0:
+            reward = torch.tensor(configs.rewardscale)
+            self.positive_reward += reward
+        self.max_end_time = self.LBs.max()
+
+        tensordict = TensorDict(
```
fedebotu, line 191: [Minor] in routing problems (e.g. [here](https://github.com/ai4co/rl4co/blob/34fe7a38080b34f0ebb643bc5f7a6fde427d065d/rl4co/envs/routing/cvrp.py#L108)), we simply use `td.update(...)` here which is slightly faster, since we don't need to re-write variables that don't change over time. 

---

NB: we make this with a "trick" in the base env, [this ](https://github.com/ai4co/rl4co/blob/34fe7a38080b34f0ebb643bc5f7a6fde427d065d/rl4co/envs/common/base.py#L96)might be useful in case it seems not to work

notebooks/unbatched_jssp.ipynb
```diff
@@ -0,0 +1,135 @@
...
+   "metadata": {
+    "collapsed": false
+   },
+   "id": "d94bd6bd3d57452a"
+  },
+  {
+   "cell_type": "code",
+   "outputs": [],
+   "source": [
+    "from torchrl.envs.vec_envs import ParallelEnv\n",
```
fedebotu, line 95: **Important**
Check out env instantiation and stepping directly on GPU, currently not working (see comment below)

fedebotu:
Closing PR for now since we have added the batched JSSP. I'd keep an eye on this though since since `ParallelEnv` can be interesting for hard-to-batch envs - @DavideTr8 if you manage to make those envs work, feel free to push a PR, looking forward to it!

PR #113, opened 2024-01-30
[Notebooks] small markdown fix in decode strategies

Haimrich:
## Description

Small fix in the markdown of the decoding strategies tutorial notebook

## Motivation and Context

Greedy and beam search subsections were inverted


fedebotu:
Thanks for spotting the typo!

PR #120, opened 2024-02-28
CVRPTW & basic Skill-VRP

ngastzepeda:
## Description

Implemented an environment for CVRPTW (Capacitated vehicle routing problem with time windows) based on the CVRP environment, extended by time windows and keeping track of the current time. Also included a basic implementation of the Skill-VRP where all technicians have the same skill, but different skill levels. Each customer node needs to be serviced by one technician, where the technician's skill-level must be >= the required skill-level at the customer node.

## Motivation and Context

The motivation is to extend the range of environments included in rl4co.

## Types of changes

- [x] Documentation (update in the documentation)

## Checklist


pyproject.toml
```diff
@@ -40,19 +40,20 @@ classifiers = [
...
     "rich",
-    "einops",
-    "wandb",
-    "matplotlib",
+    "robust-downloader",
     "scipy",
-    "robust-downloader"
+    "tensordict>=0.2.0",
+    "torchrl>=0.2.0",
+    "vrplib>=1.3.0",
```
fedebotu, line 55: (minor) should we include problem-specific dependencies as requirements, or is it better to have them as `optional-dependencies`? 
CC @cbhua

data/solomon/instances/C101.txt
fedebotu, line 1: Is there a way to have data be downloaded automatically? I think `vrplib` offered this right?

data/solomon/solutions/C101.sol
fedebotu, line 1: (same as previous comment)

rl4co/data/utils.py
```diff
@@ -21,3 +25,28 @@ def check_extension(filename, extension=".npz"):
...
+    )
+
+
+def load_solomon_solution(name, path=None):
+    if not path:
+        path = "data/solomon/solutions/"
+        path = os.path.join(ROOT_PATH, path)
+    file_path = f"{path}{name}.sol"
+    if not os.path.isfile(file_path):
+        vrplib.download_solution(name=name, path=path)
```
fedebotu, line 51: This is the way to download, then can we remove the extra `sol` and `txt` file under data? (we could simply call `os.makedirs` to generate the path for downloading automatically)

data/solomon/instances/C101.txt
fedebotu, line 1: See comment from `vrplib`, it might be better (or at least cleaner) to have problem-specific things not necessarily downloaded

pyproject.toml
fedebotu, line 55: (minor) should we include problem-specific dependencies as requirements, or is it better to have them as optional-dependencies?
CC @cbhua

rl4co/envs/routing/svrp.py
```diff
@@ -0,0 +1,402 @@
...
+                xy=(to_loc[0], to_loc[1]),
+                xytext=(from_loc[0], from_loc[1]),
+                arrowprops=dict(arrowstyle="-|>", color=out(color_idx)),
+                size=15,
+                annotation_clip=False,
+            )
+        plt.show()
+
+    @staticmethod
+    def load_data(fpath, batch_size=...):
```
fedebotu, line 400: This is already in the EnvBase class [here](https://github.com/ai4co/rl4co/blob/1d487334dbf0b2d7082cc95a95f3d33540d61109/rl4co/envs/common/base.py#L234) so it is not needed I guess?

fedebotu (commented):
Awesome job! I left some mostly minor comments here and there. Other than that, I think ready to commit!

rl4co/envs/routing/svrp.py
ngastzepeda: You are right! I had originally thought to do something similar as in the cvrptw, but there are no Solomon instances for the Skill VRP, so I didn't in the end^^ Should I just delete the function?

rl4co/data/utils.py
ngastzepeda: You are right! We could check if the path exists (which it won't the first time) as additional check before checking for the file itself and just create the path if it doesn't. Do you want me to change it?

pyproject.toml
ngastzepeda: Just let me know if I should make any changes :)

cbhua: I would prefer to set these dependencies to optional. It's better to have fewer dependencies for the main RL4CO.  Also in the future, we will have more environments from various fields. They may bring quite a few dependencies.  We don't want these to be massive in the main RL4CO. 

In the meantime, we should write a clear notice to remind users to install problem-specific dependencies. In documents or environments README/comments.

rl4co/data/utils.py
N-Wouda: Heya! Current versions of `vrplib` support this downloading functionality, we consider this somewhat deprecated and plan to remove it at some point (probably for `vrplib` 2.0.0). That should not be a problem since you pin your dependencies, but could be something to keep in mind!

fedebotu: Thanks a lot @N-Wouda for your heads up! Do you have any suggestions for alternatives?

N-Wouda: @fedebotu what we do with PyVRP is that we have our own repository of benchmark instances and best-known solutions, here: https://github.com/PyVRP/Instances. The CVRP and VRPTW instances track those in CVRPLIB, but it's nice to have everything in one place, and one format. We include this as a git submodule in the main PyVRP repo, which makes for easy benchmarking.

For RL4CO that's probably not a great option, since routing isn't all you do. Maybe point users to either our instances, or to CVRPLIB, and ask them to download the files themselves? That could be part of your docs. Alternatively, you could implement the downloading yourself: it is [only a few lines of code](https://github.com/PyVRP/VRPLIB/blob/main/vrplib/download/download_solution.py).

Hope that helps!

fedebotu: @N-Wouda thanks a ton for the detailed response! I think, as you suggested, it might be easier to just provide the utils to download in a few lines of code and then include some `vrplib` optional dependency as we are doing now for loading them :)

PR #140, opened 2024-03-15
[Feat] Add Local Search for Solution Improvement

hyeok9855:
## Description

Add local search operators as a post-processing to improve a given solution.
Here, we implement [2-opt for TSP](https://github.com/henry-yeh/DeepACO/blob/main/tsp_nls/two_opt.py) and [LocalSearch operator provided by PyVRP](https://pyvrp.org/api/search.html#module-pyvrp.search.LocalSearch) for CVRP.
For other problems (e.g., PDP or scheduling), we couldn't find such a plug-and-play local search operator, and we're looking for contributions to local search for other problems!

Note that I also made some refactorings regarding type hinting.

## Motivation and Context

Local search is an essential component for an enhanced CO solver. Many research projects in the NCO field utilize local search operators, such as our recent work [GFACS](https://github.com/ai4co/gfacs), which trains NN using the solution refined by local search in an off-policy manner.

- [ ] I have raised an issue to propose this change ([required](https://github.com/ai4co/rl4co/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Remove all that do not apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of examples)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #142, opened 2024-03-18
[Feat] Add DeepACO

Furffico:
## Description

Implementing the non-autoregressive algorithm [DeepACO](https://arxiv.org/abs/2309.14032) on RL4CO.

## Motivation and Context

As a non-autoregressive algorithm, DeepACO leverages Ant Colony Optimization (ACO) algorithms to construct solutions without relying on autoregressive decoding, thereby extending the capabilities of RL4CO.

## Types of changes

## Checklist

- [x] I have updated the documentation accordingly.

### DIFF
--- a/rl4co/models/zoo/deepaco/antsystem.py
+++ b/rl4co/models/zoo/deepaco/antsystem.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from typing import Optional
 
 import torch
@@ -25,7 +26,7 @@ class AntSystem:
         pheromone: Optional[torch.Tensor] = None,
         require_logp: bool = False,
     ):
-        self.batch_size, self.n_nodes, _ = log_heuristic.shape
+        self.batch_size = log_heuristic.shape[0]
         self.n_ants = n_ants
         self.n_iterations = n_iterations
         self.alpha = alpha
@@ -42,9 +43,6 @@ class AntSystem:
         self.all_records = []
 
         self._batchindex = torch.arange(self.batch_size, device=log_heuristic.device)
-        self._batch_n_indices = (
-            self._batchindex.unsqueeze(1).repeat(1, self.n_nodes).view(-1)
-        )
 
     def run(
         self,
@@ -57,9 +55,10 @@ class AntSystem:
         for _ in range(n_iterations):
             # reset environment
             td: TensorDict = env.reset(td_initial.clone(recurse=False))  # type: ignore
-            reward, actions = self.one_step(td, env)
+            self.one_step(td, env)
 
-        return td, actions, reward
+        td, env = self._recreate_final_routes(td_initial, env)
+        return td, self.final_actions, self.final_reward
 
     def one_step(self, td: TensorDict, env: RL4COEnvBase):
         # sampling
@@ -93,7 +92,7 @@ class AntSystem:
         reward = env.get_reward(td, actions)
 
         if self.require_logp:
-            self.all_records.append((outputs, actions, reward))
+            self.all_records.append((outputs, actions, reward, td.get("mask", None)))
 
         # reshape from (batch_size * n_ants, ...) to (batch_size, n_ants, ...)
         reward = reward.view(self.batch_size, self.n_ants)
@@ -126,13 +125,16 @@ class AntSystem:
         from_node = actions
         to_node = torch.roll(from_node, 1, -1)
         mapped_reward = self._reward_map(reward).detach()
+        batch_action_indices = self._batch_action_indices(
+            self.batch_size, actions.shape[-1], reward.device
+        )
 
         for ant_index in range(self.n_ants):
             delta_pheromone[
-                self._batch_n_indices,
+                batch_action_indices,
                 from_node[:, ant_index].flatten(),
                 to_node[:, ant_index].flatten(),
-            ] += mapped_reward[self._batch_n_indices, ant_index]
+            ] += mapped_reward[batch_action_indices, ant_index]
 
         # decay & update
         self.pheromone *= self.decay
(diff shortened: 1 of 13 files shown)


rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -0,0 +1,245 @@
...
+        if self.require_logp:
+            self.all_records.append((outputs, actions, reward, td.get("mask", None)))
+
+        # reshape from (batch_size * n_ants, ...) to (batch_size, n_ants, ...)
+        reward = reward.view(self.batch_size, self.n_ants)
+        actions = actions.view(self.batch_size, self.n_ants, -1)
+
+        return actions, reward
+
+    def _local_search(self, actions: Tensor, reward: Tensor) -> Tuple[Tensor, Tensor]:
```
fedebotu, line 135: Note: the current plan is to extend `_local_search` into an (optional) plug-in class method from the `Env` so we can fully modularize this part, which is the work  @hyeok9855 is doing as you can see in [this PR](https://github.com/ai4co/rl4co/pull/140) 

The signature would be something like:

```python
class CustomRL4COEnv()
   [...]
    def local_search(self, td: TensorDict, actions: torch.Tensor, **kwargs) -> torch.Tensor:
    """Input: td (state), actions
         Returns: improved actions
     """
```

I think this could be further modularized. For example, we could set a parameter for setting the local search operator (by default `None`). What do you think?

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -0,0 +1,245 @@
...
+                from_node[:, ant_index].flatten(),
+                to_node[:, ant_index].flatten(),
+            ] += mapped_reward[batch_action_indices, ant_index]
+
+        # decay & update
+        self.pheromone *= self.decay
+        self.pheromone += delta_pheromone
+
+    def _reward_map(self, x):
+        # map reward from $\mathbb{R}$ to $\mathbb{R}^+$
```
fedebotu, line 191: [Minor, no need to modify now] For documentation purposes in ReadTheDocs, math can be done

```python
   def _reward_map(self, x):
        """Map reward :math:`f: \mathbb{R} \rightarrow \mathbb{R}^+`"""
```

rl4co/models/zoo/deepaco/decoder.py
```diff
@@ -0,0 +1,102 @@
...
+
+        aco = self.aco_class(heuristic_logp, n_ants=self.n_ants[phase], **self.aco_args)
+        td, actions, reward = aco.run(td_initial, env, self.n_iterations[phase])
+
+        if calc_reward:
+            td.set("reward", reward)
+
+        return None, actions, td
+
+    def _conv_params(self, value: Union[int, dict, None] = None, **defaults):
```
fedebotu, line 95: I like this function! 
We should use something similar for default dicts more. Maybe it can go in `utils` somewhere?
(Also, minor thing) this is a `@staticmethod`)

fedebotu (commented):
Great job!! :rocket:

rl4co/models/zoo/deepaco/antsystem.py
Furffico: I think that would be a great idea!

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -0,0 +1,245 @@
...
+        self,
+        log_heuristic: Tensor,
+        n_ants: int = 20,
+        alpha: float = 1.0,
+        beta: float = 1.0,
+        decay: float = 0.95,
+        pheromone: Optional[Tensor] = None,
+        require_logp: bool = False,
+    ):
+        self.batch_size = log_heuristic.shape[0]
```
cbhua, line 39: [Very Minor, No need to fix now] theoretically we support multi `batch_size` dimensions. But this rarely happens  and we have to touch a few other functions to fully support that. I think it's fine to keep this.

cbhua (approved):
Awesome code! 🚀 You are really good at quality coding!

rl4co/models/zoo/deepaco/decoder.py
Furffico: Which file should I put this function in, `rl4co/utils/utils.py`? And please suggest a name for the extracted function. 😄

fedebotu: The location is good! For the name, GPT4 suggests merge_with_defaults, I guess this could be a good name

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -93,9 +96,9 @@ def one_step(self, td: TensorDict, env: RL4COEnvBase):
             reward: The reward achieved by the algorithm.
         """
         # sampling
-        actions, reward = self._sampling(td, env)
+        td, env, actions, reward = self._sampling(td, env)
         # local search, reserved for extensions
-        actions, reward = self._local_search(actions, reward)
+        actions, reward = self._local_search(td, env, actions, reward)
```
fedebotu, line 101: [Minor] in case we don't want to do local search, I think we could set a parameter for DeepACO such as `use_local_search` (I would say set to "False" by default) ; then once the `env` class has the local search option, the env could raise an assertion error if it is not implemented (e..g [here](https://github.com/ai4co/rl4co/pull/140/files#r1528357231))

fedebotu (approved):

fedebotu:
LGTM! Thoughts @cbhua?

Now the remaining things to do are:
1. Reproduce results from the paper (at least make sure it works the same as the original implementation in terms of runtime and convergence)
2. Integration with local search CC: @hyeok9855   (and relative checking results~)

cbhua:
LGTM!

PR #145, opened 2024-03-20
MatNet implementation for FFSP

LTluttmann:
## Description

- FFSPEnv:
	- bug fixes:
		- .view() on run_time messed up the run times (interleaving instead of stacking behavior)
		- reward missed the "-" sign
	- added inline comments
	- implemented render function
	- implemented a mechanism allowing for N POMO starts. Other than the original implementation, this works for arbitrary instances (instead of assuming 3 stages with 4 machines each)
	- implemented a "flatten" option, which flattens out the stage dimension. This allows the use of the (single-stage) MatNet model
	- updated MatNet test

- MatNet
	- implemented the multi-stage MatNet Model for the FFSP as described in the original MatNet paper
	- implemented a single-stage MatNet model for the FFSP. This has the advantage over the original implementation that it is size-agnostic. The multi-stage model can only be applied to instances with a fixed number of stages. The performance of the single-stage model is still under evaluation.
	- added function which, based on the environment, selects the correct MatNet policy

- MatNetCrossMHA:
	- changed some default parameters in accordance to the official paper implementation

- LogitAttention:
	- changed the keyword argument "linear_bias" to "out_bias", since "linear_bias" is already used in AutoregressiveDecoder. Thus, LogitAttention's linear_bias could not be changed without changing that of the AutoregressiveDecoder.

- nn.env_embeddings:
	- added context, init and dynamic emebddings for FFSP
	- replaced the MatNetATSPInitEmbedding with a more generic MatNetInitMebedding (works for ATSP and FFSP)

- POMO: 
	- added "max_reward" to td also during training (for logging purposes)

- RL4COEnvBase:
	- added get_num_starts and select_start_nodes as functions to the environment. By default, the get_num_starts select_start_nodes from the rl4co.utils.ops module are called (w.i.p.), but environments can override these functions with custom logic (example in FFSPEnv).
	- removed select_start_nodes_fn attribute from AutoregressiveDecoder. Note, that it can still be manually set by passing the select_start_nodes_fn argument to the forward function of the decoder (which hands it over to the DecodingStrategy)

- config (main.yaml):
	- moved "metrics" under the model config

## Motivation and Context

- Added working FFSP MatNet implementation
- improvements and fixes for FFSP environment
- let get_num_starts and select_start_nodes functions be implemented by environments as discussed here: https://github.com/ai4co/rl4co/discussions/117
- some minor fixes

## Types of changes

## Checklist


rl4co/envs/common/base.py
```diff
@@ -173,6 +174,12 @@ def get_action_mask(self, td: TensorDict) -> torch.Tensor:
         """
         raise NotImplementedError
 
+    def get_num_starts(self, td):
+        return get_num_starts(td, self.name)
```
fedebotu, line 178: Awesome! 

At the moment this is not yet fully modularized (I guess each env should implement this function)
- In the decoding, i.e. [here](https://github.com/ai4co/rl4co/blob/66622a2e7d616e72696904e1552cf7d61dad817f/rl4co/models/nn/dec_strategies.py#L70) and [here](https://github.com/ai4co/rl4co/blob/66622a2e7d616e72696904e1552cf7d61dad817f/rl4co/models/nn/dec_strategies.py#L82) I guess we should eventually call instead `env.get_num_starts` and `env.select_start_nodes` instead of having to go through the default function from `ops` (i guess this is what you meant in the  discussion!)
- [Minor] we should also add some docstring
CC: https://github.com/ai4co/rl4co/discussions/117

rl4co/envs/scheduling/ffsp.py
```diff
@@ -333,11 +421,18 @@ def generate_data(self, batch_size) -> TensorDict:
...
+            low=self.min_time,
+            high=self.max_time,
+            size=(*batch_size, self.num_job, self.num_machine, self.num_stage),
+        ).to(self.device)
+
+        if self.flatten_stages:
+            run_time = (
+                run_time.transpose(-2, -1)
+                .contiguous()
+                .view(*batch_size, self.num_job, self.num_machine_total)
```
fedebotu, line 434: [Comment only] Great, these `view` parts are always tricky... I wish there was a standardized way to do this

rl4co/envs/scheduling/ffsp.py
```diff
@@ -346,5 +441,105 @@ def generate_data(self, batch_size) -> TensorDict:
             batch_size=batch_size,
         )
 
-    def render(self, td: TensorDict):
-        raise NotImplementedError("TODO: render is not implemented yet")
+    def render(self, td: TensorDict, idx: int):
```
fedebotu, line 444: [Comment only] awesome!

rl4co/models/nn/dec_strategies.py
```diff
@@ -79,7 +77,10 @@ def pre_decoder_hook(self, td: TensorDict, env: RL4COEnvBase):
 
         # Multi-start decoding: first action is chosen by ad-hoc node selection
         if self.num_starts > 1:
-            action = self.select_start_nodes_fn(td, env, self.num_starts)
+            if self.select_start_nodes_fn is not None:
+                action = self.select_start_nodes_fn(td, env, self.num_starts)
+            else:
+                action = env.select_start_nodes(td, num_starts=self.num_starts)
```
fedebotu, line 83: [Comment only] ok, I am dumb, I commented above on something you did here already :laughing:

rl4co/models/zoo/matnet/encoder.py
```diff
@@ -51,7 +54,7 @@ def __init__(
         self.mix_W2 = nn.Parameter(mix_W2)
         self.mix_b2 = nn.Parameter(mix_b2)
 
-        self.out_proj = nn.Linear(embedding_dim, embedding_dim, bias=bias)
+        self.out_proj = nn.Linear(embedding_dim, embedding_dim, bias=True)
```
fedebotu, line 57: Is there a reason why the bias is set to True? Because of the original code?

rl4co/models/zoo/matnet/policy.py
```diff
@@ -26,36 +41,171 @@ class MatNetPolicy(AutoregressivePolicy):
...
         **kwargs,
     ):
-        if env_name not in ["atsp"]:
-            log.error(f"env_name {env_name} is not originally implemented in MatNet")
+        if env.name not in ["atsp", "ffsp"]:
+            log.error(f"env_name {env.name} is not originally implemented in MatNet")
+
+        if env.name == "ffsp":
+            decoder = MatNetFFSPDecoder(
+                env=env,
```
fedebotu, line 59: Is passing the full `env` necessary here? It seems that in the `FFSPDecoder` no env is needed

rl4co/models/zoo/matnet/policy.py
```diff
@@ -26,36 +41,171 @@ class MatNetPolicy(AutoregressivePolicy):
 
     def __init__(
         self,
-        env_name: str,
+        env: RL4COEnvBase,
```
fedebotu, line 44: [Minor] usally we allow to pass both `str` and `RL4COEnvBase` as seen [here](https://github.com/ai4co/rl4co/blob/66622a2e7d616e72696904e1552cf7d61dad817f/rl4co/models/zoo/common/autoregressive/decoder.py#L75C1-L76C1)

fedebotu (commented):
Great job :rocket:  left some comments here and there

rl4co/models/zoo/matnet/encoder.py
LTluttmann: oh yes, after struggling to obtain the results reported in the paper I tried to copy the architecture as much as possible, realizing that the bias in the projection of the heads does not have much effect. I would therefore revert this change.

rl4co/models/zoo/matnet/policy.py
LTluttmann: you are right! I will remove this line

rl4co/models/zoo/matnet/encoder.py
fedebotu: As you wish, this is minor!

fedebotu (approved):

PR #152, opened 2024-04-05
[Feat] Decoding refactoring

fedebotu:
## Description

Several refactorings and features made for decoding in RL4CO.

1. [Refactoring] Now models return logits by default (e.g. [here](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/nn/attention.py#L177)). We do so since logits represent the raw outputs from the model, and we would like to decouple the modeling part to how we sample distributions. The function handling the transfer from logits to probabilities (hence the "log_p") is [`logit_to_probs`](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/nn/dec_strategies.py#L53) 
2. [Feat] New decoding strategy: we introduce [nucleus sampling](https://arxiv.org/abs/1904.09751) (i.e. top-p sampling) which discards from the distribution values under a certain threshold in the CDF before sampling. This can be used by simply passing a `top_p`  > 0 to the `DecodingStrategy`, i.e. to the model deooder. This is ubiquitous in LLMs and it is about time to have it!
3. [Refactoring]: for simplicity we now default to **handling probabilities instead of log probabilities** (example [here](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/nn/dec_strategies.py#L202C10-L202C11)). This is a minor change, but it can make the code more readable and avoid having to do `logp.exp()` when sampling. This is also more in line with recent works in e.g. LLM
4. [Refactoring, **breaking change**] now by default **any mask has the same behavior** (example [here](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/nn/dec_strategies.py#L72)), i.e., the value 1 means keep (i.e. feasible action) while 0 means to remove, i.e. infeasible. This is both similar to TorchRL's action mask and importantly to PyTorch's `scaled_dot_product_attention`: "A boolean mask where a value of True indicates that the element should take part in attention. " ([ref](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#:~:text=A%20boolean%20mask%20where%20a%20value%20of%20True%20indicates%20that%20the%20element%20should%20take%20part%20in%20attention.)). For this reason, masks that used to have inconsistent namings now have the same behavior
5. [Minor] Rename `LogitAttention` to [`PointerAttention`](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/nn/attention.py#L133) (for consistency with the Pointer mechanism in [Vinyals et al., 2015](https://arxiv.org/abs/1506.03134))

> [!WARNING]
> Work in progress. Do not merge yet. Some checks and training still have some bugs that need to be fixed (most probably due to the new masking

## Types of changes

## Checklist

---

CC: @LTluttmann could you have a look if you spot some inefficiencies or if you have some ideas?
CC: @Furffico @cbhua these changess are what I was talking about yesterday (note that in this case running the softmax normalization inside the[ `Sampling` ](https://github.com/ai4co/rl4co/blob/2c6984f4d9bd68252341347abd6c2cb0f80fdf27/rl4co/models/zoo/deepaco/antsystem.py#L124)in ACO might not be needed)

### DIFF
--- a/configs/experiment/routing/am-xl.yaml
+++ b/configs/experiment/routing/am-xl.yaml
@@ -14,7 +14,7 @@ logger:
   wandb:
     project: "rl4co"
     tags: ["am", "${env.name}"]
-    group: ${env.name}${env.num_loc}"
+    group: "${env.name}${env.num_loc}"
     name: "am-xl-${env.name}${env.num_loc}"
 
 model:
--- a/configs/experiment/routing/pomo.yaml
+++ b/configs/experiment/routing/pomo.yaml
@@ -14,7 +14,7 @@ logger:
   wandb:
     project: "rl4co"
     tags: ["pomo", "${env.name}"]
-    group: ${env.name}${env.num_loc}"
+    group: "${env.name}${env.num_loc}"
     name: "pomo-${env.name}${env.num_loc}"
 
 model:
--- a/configs/experiment/routing/symnco.yaml
+++ b/configs/experiment/routing/symnco.yaml
@@ -14,7 +14,7 @@ logger:
   wandb:
     project: "rl4co"
     tags: ["symnco", "${env.name}"]
-    group: ${env.name}${env.num_loc}"
+    group: "${env.name}${env.num_loc}"
     name: "symnco-${env.name}${env.num_loc}"
 
 model:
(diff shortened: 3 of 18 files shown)


fedebotu:
Closed in favor of #161

PR #153, opened 2024-04-08
Create __init__.py

bokveizen:
Added the missing `__init__.py` under https://github.com/ai4co/rl4co/tree/main/rl4co/models/zoo/equity_transformer


fedebotu:
About [Equity Transformer](https://arxiv.org/abs/2306.02689): we can simply put the `init` and `context` embeddings instead since the model is the same as AM. Let's merge for now though :+1: 

CC: @Leaveson 

(same applies for other PRs)

PR #159, opened 2024-04-09
Update edge.py

bokveizen:
edge_attr should have the same length as the number of edges not nodes

### DIFF
--- a/rl4co/models/nn/env_embeddings/edge.py
+++ b/rl4co/models/nn/env_embeddings/edge.py
@@ -144,7 +144,7 @@ class NoEdgeEmbedding(nn.Module):
             data = Data(
                 x=node_embed,
                 edge_index=edge_index,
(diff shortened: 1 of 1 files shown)


fedebotu (commented):
I think `n` was defined just few lines above, so in this case we should delete that as well?
This said it seems correct to me, could you verify @Furffico ?

bokveizen:
> I think `n` was defined just few lines above, so in this case we should delete that as well? This said it seems correct to me, could you verify @Furffico ?

`n` is used for `get_full_graph_edge_index` though.

Furffico:
> I think `n` was defined just few lines above, so in this case we should delete that as well? This said it seems correct to me, could you verify @Furffico ?

Yes, it is correct for `edge_attr` to have the same length as the number of edges.

fedebotu:
LGTM!

PR #161, opened 2024-04-09
[Feat, Refactor] Decoding: logits handling, new decoding strategies, standardize masks

fedebotu:
## Description

Several refactorings and features made for decoding in RL4CO.
Note that this PR follows #152 but restructured - now working and tested!

1. [Refactoring] Now models return logits by default (e.g. [here](https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/models/nn/attention.py#L201)). We do so since logits represent the raw outputs from the model, and we would like to decouple the modeling part to how we sample distributions. The function handling the transfer from logits to log probabilities (hence the "log_p") is [`process_logits`](https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/utils/decoding.py#L136) 
2. [Feat] New decoding strategies!
    - We introduce [nucleus sampling](https://arxiv.org/abs/1904.09751) (i.e. top-p sampling) which discards from the distribution values under a certain threshold in the CDF before sampling. This can be used by simply passing a `top_p`  > 0 to the `DecodingStrategy`, i.e. to the model deooder. This is ubiquitous in LLMs and it is about time to have it!
    - Similarly we introduce top-k sampling which considers simply the top-k highest logits in the distribution to sample from
3. [Refactoring, **breaking change**] now by default **any mask has the same behavior** (example [here](https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/utils/decoding.py#L155)), i.e., the value 1 means keep (i.e. feasible action) while 0 means to remove, i.e. infeasible. This is both similar to TorchRL's action mask and importantly to PyTorch's `scaled_dot_product_attention`: "A boolean mask where a value of True indicates that the element should take part in attention. " ([ref](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#:~:text=A%20boolean%20mask%20where%20a%20value%20of%20True%20indicates%20that%20the%20element%20should%20take%20part%20in%20attention.)). For this reason, masks that used to have inconsistent namings now have the same behavior
4. [Minor] Rename `LogitAttention` to [`PointerAttention`](https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/models/nn/attention.py#L133) (for consistency with the Pointer mechanism in [Vinyals et al., 2015](https://arxiv.org/abs/1506.03134))
5. Transfer all decoding utilities (such as decoding strategies, log likelihood calculation and so on) under `rl4co/utils` [here](https://github.com/ai4co/rl4co/blob/refactor-decoding-new/rl4co/utils/decoding.py)

## Types of changes

## Checklist

---

CC: @LTluttmann could you have a look if you spot some inefficiencies or if you have some ideas?
CC: @Furffico @cbhua (note: made several modifications compared to previous PR)


rl4co/models/zoo/mdam/decoder.py
```diff
@@ -85,8 +86,9 @@ def __init__(
             env_name, {"embedding_dim": embedding_dim}
         )
 
-        self.logit_attention = [
-            LogitAttention(
+        # MHA with Pointer mechanism (https://arxiv.org/abs/1506.03134)
+        self.pointer = [
```
LTluttmann, line 90: seems like the pointers in self.pointer are never used. Probably, they should replace the "_one_to_many_logits" function?

rl4co/models/zoo/ptrnet/decoder.py
```diff
@@ -157,7 +157,8 @@ def forward(
             )
             # select the next inputs for the decoder [batch_size x hidden_dim]
             idxs = (
-                decode_probs(probs, mask, decode_type=decode_type)
+                # note: mask here is the inverse of usual action mask
+                decode_logprobs(log_p, ~mask, decode_type=decode_type)
```
LTluttmann, line 161: maybe change the masking here to be consistent with the rework, i.e.:
- start with a attn_mask of ones (i.e. use torch.ones) here:
https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/models/zoo/ptrnet/decoder.py#L147
- gradually set selected actions to False here: https://github.com/ai4co/rl4co/blob/8e9595bf636f5ba5106811a57d58eadd9e255132/rl4co/models/zoo/ptrnet/decoder.py#L75

rl4co/models/zoo/common/nonautoregressive/decoder.py
```diff
@@ -156,33 +156,27 @@ def forward(
 
         # Main decoding: loop until all sequences are done
         while not td["done"].all():
-            log_p, mask = self._get_log_p(td, heatmaps_logp, num_starts)
-            td = decode_strategy.step(log_p, mask, td)
+            logits, mask = self._get_logits(td, heatmaps_logits, num_starts)
+            td = decode_strategy.step(logits, mask, td)
```
fedebotu, line 160: Not 100% sure about this, since you use the sigmoid actually we may not need to  normalize the "logits". Or do we? @Furffico

rl4co/models/zoo/mdam/decoder.py
```diff
@@ -272,15 +274,15 @@ def _get_log_p(self, fixed, td, path_index, normalize=True):
         glimpse_v = fixed.glimpse_val + glimpse_val_dynamic
         logit_k = fixed.logit_key + logit_key_dynamic
 
-        # Compute the mask
+        # Compute the mask (NOTE: here it is the opposite due to custom attention implementation)
         mask = ~td["action_mask"]
```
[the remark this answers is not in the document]
fedebotu: You're right, I think I skipped this due to lazyness but it can be further modularized. It seems no strange attention implementation was used in this model anyways!

rl4co/utils/decoding.py
```diff
@@ -0,0 +1,502 @@
...
+    # Compute log probabilities
+    return F.log_softmax(logits, dim=-1)
+
+
+class DecodingStrategy:
+    """Base class for decoding strategies. Subclasses should implement the :meth:`_step` method.
+    Includes hooks for pre and post main decoding operations.
+
+    Args:
+        temperature (float, optional): Temperature scaling. Higher values make the distribution more uniform (exploration),
```
fedebotu, line 194: [Minor] use consistent typing with the `typing` module

rl4co/utils/decoding.py
```diff
@@ -0,0 +1,502 @@
...
+    # First sort and calculate cumulative sum of probabilities.
+    sorted_logits, sorted_indices = torch.sort(logits, descending=False)
+    cumulative_probs = sorted_logits.softmax(dim=-1).cumsum(dim=-1)
+
+    # Remove tokens with cumulative top_p above the threshold (token with 0 are kept)
+    sorted_indices_to_remove = cumulative_probs <= (1 - top_p)
+
+    # scatter sorted tensors to original indexing
+    indices_to_remove = sorted_indices_to_remove.scatter(
+        1, sorted_indices, sorted_indices_to_remove
```
fedebotu, line 131: Note: this should be `-1` for multi-dimensional tensors

rl4co/utils/decoding.py
```diff
@@ -0,0 +1,502 @@
...
+
+    @staticmethod
+    def greedy(logprobs, mask=None):
+        """Select the action with the highest probability."""
+        # [BS], [BS]
+        selected = logprobs.argmax(dim=-1)
+        if mask is not None:
+            assert (
+                not (~mask).gather(1, selected.unsqueeze(-1)).data.any()
+            ), "infeasible action selected"
```
Furffico, line 321: Is it still required to check here (also in sampling, L331-337) as the logprobs are already masked in `process_logits` (L298-306)?

rl4co/models/zoo/common/nonautoregressive/decoder.py
Furffico: > Not 100% sure about this, since you use the sigmoid actually we may not need to normalize the "logits". Or do we? @Furffico

That's correct, we don't need to normalize the logits in this version. I tested and it worked fine.

Furffico (approved):
Great job! :tada: I have some small suggestions in the new `decoding.py`.

rl4co/models/zoo/mdam/decoder.py
fedebotu: (As discussed, left as TODO for the future for the time being!)

rl4co/models/zoo/mdam/decoder.py
```diff
@@ -254,7 +256,7 @@ def _make_heads(self, v, num_steps=None):
             )  # (n_heads, batch_size, num_steps, graph_size, head_dim)
         )
 
-    def _get_log_p(self, fixed, td, path_index, normalize=True):
+    def _get_logprobs(self, fixed, td, path_index, normalize=True):
```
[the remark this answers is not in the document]
fedebotu: (As discussed, left as TODO for the future for the time being!)

rl4co/models/zoo/mdam/decoder.py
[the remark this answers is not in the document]
fedebotu: (As discussed, left as TODO for the future for the time being!)

rl4co/models/zoo/mdam/decoder.py
cbhua: Yes I think this kind of redundancy may come from the transferring from the original [MDAM implementation](https://github.com/liangxinedu/MDAM/blob/19b0bf813fb2dbec2fcde9e22eb50e04675400cd/nets/model_search.py#L51). Let me check it then.

rl4co/models/zoo/ptrnet/decoder.py
fedebotu: Done!

rl4co/models/zoo/common/nonautoregressive/decoder.py
fedebotu: I see - but then isn't it better to use `log_softmax` on the heatmap rows rather than normalizing each element with sigmoid? 

Anyways this version should give the same solution in terms of sampling and greedy but it also supports the new features. So if you agree we could leave it as is?

rl4co/utils/decoding.py
fedebotu: It can be needed if one passes a custom function that does not mask the logprobs (e.g. if `sampling` and `greedy` are called directly)

rl4co/utils/decoding.py
```diff
@@ -0,0 +1,502 @@
...
+    """
+
+    # Tanh clipping from Bello et al. 2016
+    if tanh_clipping > 0:
+        logits = torch.tanh(logits) * tanh_clipping
+
+    # In RL, we want to mask the logits to prevent the agent from selecting infeasible actions
+    if mask_logits:
+        assert mask is not None, "mask must be provided if mask_logits is True"
+        logits[~mask] = float("-inf")
```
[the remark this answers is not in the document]
fedebotu: I agree, however we left it because some models call `process_logits` directly and may need this for safety checking

rl4co/models/nn/attention.py
```diff
@@ -129,97 +130,104 @@ def forward(self, x, key_padding_mask=None):
...
         k = self._make_heads(key)
         v = self._make_heads(value)
-
         if self.mask_inner:
-            # need to invert mask: (N L S) -> (N 1 L S)
+            # make mask the same number of dimensions as q
             attn_mask = (
-                ~mask.unsqueeze(1) if mask.ndim == 3 else ~mask.unsqueeze(1).unsqueeze(2)
+                attn_mask.unsqueeze(1)
+                if attn_mask.ndim == 3
```
cbhua, line 211: Since we are considering multi batch dimensions in this version, should this be modified?

cbhua (approved):
Awesome refractory! 🚀 I will try to create an UML graph for the overview pipeline. Could be helpful for users to understand.

rl4co/models/nn/attention.py
fedebotu: Actually for the `batchify` refactoring we considered it might take a bit more time and effort (we planned to support _any_ leading batch dimension for all operations, which right now is not supported since we do reshaping via `batchify` and `unbatchify`).
I think still this is a bit low in the priority list since it turns out many libraries do not consider multiple batch dimensions - PyTorch's own [MultiheadAttention](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html) does not, for instance

fedebotu:
> Awesome refractory! 🚀 I will try to create an UML graph for the overview pipeline. Could be helpful for users to understand.

Merging now!

PR #164, opened 2024-04-18
Update four new Envs

FeiLiu36:
Update four new Envs (VRPTW, OVRP, VRPL, VRPB)


fedebotu (commented):
I left some comments~

Some questions and notes:

1. In [your paper](https://arxiv.org/abs/2402.16891), did you implement an environment for each attribute composition? In your case, it seems there should be 16 different problems. Did you change the environment at each iteration to change the task?  Or did you make a single environment that can handle any constraint?

2. About backhauls: there may be a problem in the calculation of backhauls [here](https://github.com/FeiLiu36/rl4co/blob/1ac45161ae6da00abd9dc5e3b34d06c04f56464c/rl4co/envs/routing/vrpb.py#L95):
```python
        # Increase capacity if depot is not visited, otherwise set to 0
        used_capacity = (td["used_capacity"] + selected_demand) * (
            current_node != 0
        ).float()
 ```
 
 In this case, backhaul nodes can be visited first and have demand < 0. The used capacity would become negative [here](https://github.com/FeiLiu36/rl4co/blob/1ac45161ae6da00abd9dc5e3b34d06c04f56464c/rl4co/envs/routing/vrpb.py#L159) but actually backhauls should be counted towards exceeding the capacity?
 
 Also, since backhauls and linehauls are described by "demand", only a single resource is considered and can be delivered from a node to another (e.g. if `used_capacity` approaches the vehicle capacity, then a backhaul node with "negative demand" would decrease it). Is this correct?

File: rl4co/utils/ops.py [hunk 1]
```diff
@@ -93,7 +93,31 @@ def get_tour_length(ordered_locs):
...
+    segment_lengths = ((ordered_locs_next-ordered_locs)**2).sum(-1).sqrt()
+
+    # Get the first value of ordered_locs
+    first_loc = ordered_locs[:, 0, :][:,None,:].expand(ordered_locs_next.shape)
+
+    # Check the ids where the location is the same as the first value
+    same_loc_ids = torch.all(ordered_locs_next == first_loc, dim=-1)
+
+    # for open VRP, the distance between last customer and the depot is not counted
+    segment_lengths[same_loc_ids] = 0
```
ngastzepeda, line 115: Is it just the last customer for which we don't need to return to the depot? Why don't you do something like the following, then:
```
def get_open_tour_length(ordered_locs):
    ordered_locs_next = torch.roll(ordered_locs, -1, dims=-2)
    return get_distance(ordered_locs[:, :-1, :], ordered_locs_next[:, :-1, :]).sum(-1)
```
Notice how (1) in torch.roll `shifts=-1` instead of 1 (i.e. we shift the sequence to the left), and (2) in `get_distance()` the order of the two parameters is also exchanged. But just omitting the last entry in both tensors you're not travelling back to the depot.

ngastzepeda: I have not actually run the code, but I did something similar (without the indexing) at a different point, so in principle it shoudl work :)

ngastzepeda (changes requested):
I would like to remove retundant code, i.e. let classes inherit from parent classes and use the parent classes parameters and functions, where possible. That makes the code cleaner and easier to maintain :)

rl4co/envs/__init__.py
```diff
@@ -18,6 +18,10 @@
     SPCTSPEnv,
     TSPEnv,
     MDCPDPEnv,
+    VRPLEnv,
+    OVRPEnv,
+    VRPTWEnv,
```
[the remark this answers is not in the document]
fedebotu: Right, since they are the same environment, we could have a simple option to pass the data generator as a user choice, but should not be different classes

rl4co/utils/ops.py
fedebotu: I think if it's the last customer for each route, we may still need to mask out the middle lengths.
I think that in your code snippet, an action sequence like `[0, 3, 5, 2, 0, 1, 4, 0]` where 0 is the depot will consider only the last route (0, 1, 4)  as open but not (0, 3, 5, 2)

ngastzepeda: Aah, ok, so all individual routes are open. Then my solution doesn't make sense :D

fedebotu: Yes, according the the definition, it seems to be the case (e.g. [here](https://www.sciencedirect.com/science/article/abs/pii/S0305054805003515#:~:text=to%20the%20depot.-,In%20the%20open%20vehicle%20routing%20problem%20(OVRP)%2C%20a%20vehicle%20does%20not%20return%20to%20the%20depot%20after%20servicing%20the%20last%20customer%20on%20a%20route.,-Each%20route%20in))

FeiLiu36:
> I left some comments~
> 
> Some questions and notes:
> 
> 1. In [your paper](https://arxiv.org/abs/2402.16891), did you implement an environment for each attribute composition? In your case, it seems there should be 16 different problems. Did you change the environment at each iteration to change the task?  Or did you make a single environment that can handle any constraint?
> 2. About backhauls: there may be a problem in the calculation of backhauls [here](https://github.com/FeiLiu36/rl4co/blob/1ac45161ae6da00abd9dc5e3b34d06c04f56464c/rl4co/envs/routing/vrpb.py#L95):
> 
> ```python
>         # Increase capacity if depot is not visited, otherwise set to 0
>         used_capacity = (td["used_capacity"] + selected_demand) * (
>             current_node != 0
>         ).float()
> ```
> 
> In this case, backhaul nodes can be visited first and have demand < 0. The used capacity would become negative [here](https://github.com/FeiLiu36/rl4co/blob/1ac45161ae6da00abd9dc5e3b34d06c04f56464c/rl4co/envs/routing/vrpb.py#L159) but actually backhauls should be counted towards exceeding the capacity?
> 
> Also, since backhauls and linehauls are described by "demand", only a single resource is considered and can be delivered from a node to another (e.g. if `used_capacity` approaches the vehicle capacity, then a backhaul node with "negative demand" would decrease it). Is this correct?

rl4co/envs/routing/ovrp.py
```diff
@@ -0,0 +1,440 @@
...
+    125: 55.0,
+    150: 60.0,
+    200: 70.0,
+    500: 100.0,
+    1000: 150.0,
+}
+
+
+class OVRPEnv(RL4COEnvBase):
+    """Capacitated Vehicle Routing Problem (CVRP) environment.
```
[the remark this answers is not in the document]
FeiLiu36: Thank you for the suggestions and refinements ~ 
1): in the multi-task learning (MTL) case (my paper) the VRP environments are implemented together in a single MT environment. I am currently exploring ways to incorporate MTL into the RL4CO flowchart. It would be more convenient if we could directly integrate MTL with the existing single-task Envs in RL4CO instead of creating separate MTL Envs for each MT combination. As the first stage, the four VRP Envs have been included in RL4CO as individual VRP Envs.

2): Yes, the VRPB settings are slightly different from traditional settings. In this case, only one resource is permitted for each node, and the total demands carried by the vehicle may rise, while the maximum delivery capacity (at the depot) remains constant. This adjustment is made for ease of use in the same RL framework with a masking procedure. We may revise the settings to meet diverse real-world cases.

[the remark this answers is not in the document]
fedebotu: 1) I see! [This](https://github.com/FeiLiu36/MTNCO/blob/main/MTPOMO/POMO/VRPEnv.py) is the environment, right? A similar procedure can be adopted in RL4CO too. I guess the single environments without feature combinations could be added as well first for clarity, while for the combination, an environment as `AnyVRP` (as we are calling it) or `MTLVRP` 

2) I see! Do you have some references about the problem variant in which one resource can be transported from a node to another?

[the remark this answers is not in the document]
FeiLiu36: 1. Yes, you are right! I would suggest MTLVRP or MTVRP if it is possible~

2. The setting can be regarded as mixed VRPB, where a single resource can serve both as a linehaul and backhaul. This is one of the pioneering works on mixed VRPB (An investigation into a new class of vehicle routing problem with backhauls. Omega, 2002.). The recent revisions of mixed VRPB and more variants can be found in this review (Vehicle routing with backhauls: Review and research perspectives. Computers & Operations Research, 2018)

PR #167, opened 2024-04-26
Update ppo.py

bokveizen:
added "load_from_checkpoint" for PPO

## Description

The method "load_from_checkpoint" was missing in PPO.

## Motivation and Context

It allows "load_from_checkpoint" of PPOModel for instances now.

## Types of changes

## Checklist

### DIFF
--- a/rl4co/models/rl/ppo/ppo.py
+++ b/rl4co/models/rl/ppo/ppo.py
@@ -1,14 +1,18 @@
-from typing import Any, Union
+from typing import IO, Any, Optional, Union, cast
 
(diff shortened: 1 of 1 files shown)


fedebotu:
Thanks! The previous version was not working?
Asking because the `load_from_checkpoint` is actually overwriting a class method of the `LightningModule` as seen [here](https://lightning.ai/docs/pytorch/stable/common/checkpointing_basic.html#lightningmodule-from-checkpoint), so if it does not cause problems, it could be left

PR #170, opened 2024-05-06
[BugFix] Fix the performance issue of DeepACO

Furffico:
## Description

This pull request fixes the performance issue of the DeepACO implementation, and makes some tiny modifications to enhance the performance.


rl4co/models/zoo/nargnn/encoder.py
```diff
@@ -71,6 +71,7 @@ def _make_heatmaps(self, batch_graph: Batch) -> Tensor:  # type: ignore
             device=device,
             dtype=graphs[0].edge_attr.dtype,
         )
+        heatmaps_logits.fill_(-10)
```
fedebotu, line 74: Why do we do this? Is it because we use the decoding strategy?

Furffico: In the sparsification stage, some long edges are ruled out from the graph to reduce computational costs of the GNN. 
In the original implementation the heuristic value of these excluded edges are set to a small value. But here we are using logits, and 0 corresponds to 1 after exponentiation. This could "spread out" the heuristics and increase the probability of selecting those long edges, which causes the performance issue.
The solution is to set the value to a negative number, and I think -10 is small enough, as the range of the GNN output is $[0,10]$.

hyeok9855: Oh, I see the reason; the logits are in [-10, +10], due to line 58. So, the logits for masked edges should be -10 to inactivate them.

hyeok9855: BTW why is the logit in that range?

hyeok9855: @fedebotu BTW, why we use sigmoid for the logit values? Wouldn't it be better to use tanh to avoid this postprocessing?

Furffico: I was thinking of keeping the network structure while implementing DeepACO in rl4co, so I didn't pay much attention to the migration from probs to logits and change the output activation function. 
I'm open to suggestions and I believe we can find a better plan for it. 😄

hyeok9855: Can we avoid this by taking logarithms of the heuristics we did for the pheromone [here](https://github.com/ai4co/rl4co/pull/170/files#diff-826642af91d93eb12e98537d3c99bfdb6f472be68518aba0dcab72b1c412791dR127)?

i.e., like this: 
`heatmaps_logits = (self.alpha * torch.log(self.pheromone) + self.beta * torch.log(self.heuristic))`

Would this cause some side effects?

hyeok9855 (commented):
Thank you for your work! I have added some comments.

rl4co/models/zoo/nargnn/encoder.py
Furffico: I tried to replace it with the original output layer as in DeepACO (`heatmaps_logits` = $\log\sigma(x)$ ), but the performance is worse than the current version (`heatmaps_logits` = $10\sigma(x)$ ). So I think there might be some advantages with the current version. 🤔

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -46,10 +46,15 @@ def __init__(
         self.beta = beta
         self.decay = decay
 
-        self.log_heuristic = log_heuristic
-        self.pheromone = (
-            torch.ones_like(log_heuristic) if pheromone is None else pheromone
-        )
+        self.log_heuristic = log_heuristic - log_heuristic.flatten(-2, -1).max(
+            -1, keepdim=True
+        )[0].unsqueeze(-1)
```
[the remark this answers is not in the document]
Furffico: Thank you for the suggestions. This part is indeed redundant.

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -46,10 +46,15 @@ def __init__(
...
+
+        if pheromone is None:
+            self.pheromone = torch.ones_like(log_heuristic)
+            self.pheromone.fill_(0.0005)
```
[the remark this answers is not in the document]
Furffico: This initialization strategy is intended to amplify the influence of the pheromone during the initial search process, as it allows for a more significant impact on the ant colony in the early rounds to accelerate the convergence, which conforms my experiment results.
Moreover, Dorigo et al. (1996) initialize the pheromone value with 0.005 in the vanilla Ant System algorithm.

Dorigo, M., Maniezzo, V., & Colorni, A. (1996). Ant system: Optimization by a colony of cooperating agents. IEEE Transactions on Systems, Man, and Cybernetics, Part B (Cybernetics), 26(1), 29–41. https://doi.org/10.1109/3477.484436

rl4co/models/zoo/nargnn/encoder.py
fedebotu: A small note on this: why do we have a scaled sigmoid $10 \sigma(x)$ on the last layer in the first place? Is it for regularization? Asking because we have the [decoding strategy](https://github.com/ai4co/rl4co/blob/c546234360e4a5e772aa8a3ae8e58e85d0698492/rl4co/utils/decoding.py#L188) that processes the logits afterwards anyways

hyeok9855: I think it's about the scale of the logits, which is related to the sharpness of the probability.
Maybe we can avoid using $10\sigma(x)$ by adjusting the temperature?

rl4co/models/zoo/deepaco/antsystem.py
```diff
@@ -198,9 +203,14 @@ def _update_pheromone(self, actions, reward):
...
-    def _reward_map(self, x):
+    def _reward_map(self, x: Tensor):
         """Map reward :math:`f: \\mathbb{R} \\rightarrow \\mathbb{R}^+`"""
-        return torch.where(x >= -2, 0.25 * x + 1, -1 / x)
+        M, _ = x.max(-1, keepdim=True)
+        mean = x.mean(-1, keepdim=True)
+        v = (M - mean) / (2 * M - x - mean)
+        v[x < mean] /= 5
+        v[x == mean] = 0.5
+        return v
```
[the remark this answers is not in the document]
hyeok9855: Thanks. I think it is reasonable.
BTW, what about just simply applying the min-max scaling, i.e., $\Delta \tau(r_i) = \frac{r_i - \min_j r_j}{\max_j r_j - \min_j r_j} \in [0, 1]$?

rl4co/models/zoo/nargnn/encoder.py
fedebotu: @hyeok9855 yes, that is also an option. $\tau$ set to 0.1 should have a similar effect

Furffico: Great idea, I'll use temperature instead.

fedebotu:
@Furffico did the above fixes make DeepACO match the performance in your paper?

Furffico:
> @Furffico did the above fixes make DeepACO match the performance in your paper?

Yes. With a parallelized AS and [a better pheromone update function](https://github.com/ai4co/rl4co/pull/170#discussion_r1598088413), it's faster and sometimes perform better than our original implementation.

fedebotu:
> Yes. With a parallelized AS and [a better pheromone update function](https://github.com/ai4co/rl4co/pull/170#discussion_r1598088413), it's faster and sometimes perform better than our original implementation.

Good! Will the same hold with local search too @hyeok9855 ?

hyeok9855 (approved):
LGTM
I will check the code more thoroughly later when I work for the local search #140 (maybe next week).
Thanks for your work!

hyeok9855:
> Good! Will the same hold with local search too @hyeok9855 ?

Yes, very likely.

Furffico:
I think there's no further modifications required in this PR. Shall we merge it now?

PR #176, opened 2024-05-13
Update MTVRP

FeiLiu36:
## Changes

+ Avoid checking the time window of the last node to the depot if it is an open route 
+ Set the max time window to 4.6 instead of inf
+ Create MTVRPContext and MTVRPInitEmbedding


rl4co/models/nn/env_embeddings/init.py
```diff
@@ -146,6 +145,28 @@ def forward(self, td):
...
+    def __init__(self, embed_dim, linear_bias=True, node_dim: int = 5):
+        # node_dim = 5: x, y, demand, tw start, tw end
+        super(MTVRPInitEmbedding, self).__init__(embed_dim, linear_bias, node_dim)
+
+    def forward(self, td):
+        depot, cities = td["locs"][:, :1, :], td["locs"][:, 1:, :]
+        #durations = td["durations"][..., 1:]
+        time_windows = td["time_windows"][..., 1:, :]
+        # embeddings
+        demands =  td["demand_linehaul"][..., None] - td["demand_backhaul"][..., None]
```
fedebotu, line 160: It makes sense; basically, if it's "-", the model will understand it is a backhaul. I was thinking about having a flag, but this is also good

rl4co/models/nn/env_embeddings/context.py
```diff
@@ -146,6 +147,50 @@ def _state_embedding(self, embeddings, td):
...
+    def __init__(self, embed_dim):
+        super(VRPBContext, self).__init__(
+            embed_dim=embed_dim, step_context_dim=embed_dim + 4
+        )
+
+    def _state_embedding(self, embeddings, td):
+
+        capacity = super()._state_embedding(embeddings, td)
+        current_time = td["current_time"]
+        current_length = td["current_route_length"]
```
fedebotu, line 187: How does the model understand whether there is a limit? 
In case there is no limit (say CVRP), then it will be the same as having VRPL, since the model does not know whether the constraint will be enforced or not

rl4co/envs/routing/mtvrp/generator.py
```diff
@@ -256,7 +256,8 @@ def _default_open(td, remove):
     @staticmethod
     def _default_time_window(td, remove):
         default_tw = torch.zeros_like(td["time_windows"])
-        default_tw[..., 1] = float("inf")
+        #default_tw[..., 1] = float("inf")
+        default_tw[..., 1] = 4.6 # max tw
```
fedebotu, line 260: Doesn't this influence the solution? If default time window is 4.6, the problem should not be a CVRP but a "relaxed" VRPTW.

The reason why I thought having "inf" is because it can generalize to any scale - for the embedding, this can be set as:
```python
time_windows = torch.nan_to_num(td["time_windows"][..., 1:, :], posinf=0.0)
```
So it shouldn't influence the calculation as describe in Section 4.1 (Attribute composition) in your paper.
What do you think?

rl4co/envs/routing/mtvrp/env.py
```diff
@@ -349,9 +349,14 @@ def check_solution_validity(td: TensorDict, actions: torch.Tensor):
             curr_time = torch.max(
                 curr_time + dist, gather_by_index(td["time_windows"], next_node)[..., 0]
             )
+
+            new_shape = curr_time.size()
+            skip_open_end = td["open_route"].view(*new_shape) & (next_node == 0).view(*new_shape)
```
fedebotu, line 354: Makes sense, good catch.
Anyways, I recommend setting `check_solution` to `False` when training; otherwise, the solution will be checked at each step and it can be a bit slow. I will add a warning

fedebotu (commented):
Awesome! Added some comments~

rl4co/envs/routing/mtvrp/env.py
ngastzepeda: Actually, I don't think this is necessary. Since `skip_open_end` will only be `true` if `next_node == 0`, and since the depot has the highest time window end, `curr_time <= gather_by_index(td["time_windows"], next_node)[..., 1])` should always be `True` *except* when `curr_time` is very close to the max time already and then the duration in that last node is long enough to go over the time limit - is that something we want to allow?

rl4co/envs/routing/mtvrp/env.py
```diff
@@ -450,7 +455,7 @@ def _make_spec(self, td_params: TensorDict):
     def check_variants(td):
         """Check if the problem has the variants"""
         has_open = td["open_route"].squeeze(-1)
-        has_tw = (td["time_windows"][:, :, 1] != float("inf")).any(-1)
+        has_tw = (td["time_windows"][:, :, 1] != 4.6).any(-1)
```
cbhua, line 458: Same as the discussion with the `_default_time_window()`, Changing this bound in the setting will need to modify this part. Any reason for this hardcode?

cbhua (commented):
Good job! 🚀

rl4co/models/nn/env_embeddings/init.py
```diff
@@ -146,6 +145,28 @@ def forward(self, td):
...
+    
+
+class MTVRPInitEmbedding(VRPInitEmbedding):
+    def __init__(self, embed_dim, linear_bias=True, node_dim: int = 5):
+        # node_dim = 5: x, y, demand, tw start, tw end
+        super(MTVRPInitEmbedding, self).__init__(embed_dim, linear_bias, node_dim)
+
+    def forward(self, td):
+        depot, cities = td["locs"][:, :1, :], td["locs"][:, 1:, :]
+        #durations = td["durations"][..., 1:]
```
ngastzepeda, line 157: Why are the durations not included?

ngastzepeda (changes requested):
I agree with @fedebotu and @cbhua that we shouldn't hardcode the duration of 4.6 here!

rl4co/models/nn/env_embeddings/init.py
FeiLiu36: A flag is also a good idea!

rl4co/models/nn/env_embeddings/context.py
FeiLiu36: You are right. The route length in the state_emdding should be the rest length, i.e., length limit-current length, instead of current length... It seems to be a mistake...

rl4co/envs/routing/mtvrp/env.py
FeiLiu36: I agree with ngastzepeda as the last node in the route should also satisfy the time window constraints (allow it back it depot even when it is OVRP). However, I find some outliners when training the MTVRP (i.e., the time window of the last node of OVRP route may exceed the max time window). I do not yet know the exact reason, instance generation, or masking procedure.

rl4co/envs/routing/mtvrp/generator.py
FeiLiu36: OK, that makes sense!

rl4co/envs/routing/mtvrp/env.py
FeiLiu36: I will have a check.

rl4co/envs/routing/mtvrp/env.py
FeiLiu36: Avoid numerical issues during training as it will go through embedding, but I will have a check, the inf would be more general.

fedebotu:
Closing as stale

PR #177, opened 2024-05-16
[Feat] Add JSSP environment

LTluttmann:
## Description

- Added environment for the Job-Shop Scheduling Problem (JSSP). This implementation of JSSP treats JSSP as a special case of FJSP, where each operation can be processed by only one machine. As such, the environment is implemented as subclass of the FJSPEnv, only changing the action space (action space reduces to the selection of next job to execute) and the data generator. 
- In addition, the HetGNN policy has been restructured and renamed to L2D (for learning to dispatch) and is now applicable to both, FJSP and JSSP. 

## Motivation and Context

- JSSP is a common CO problem and a widely used problem to benchmark new algorithms

## Types of changes

## Checklist


rl4co/models/__init__.py
```diff
@@ -19,7 +19,7 @@
 from rl4co.models.rl.ppo.ppo import PPO
 from rl4co.models.rl.reinforce.baselines import REINFORCEBaseline, get_reinforce_baseline
 from rl4co.models.rl.reinforce.reinforce import REINFORCE
-from rl4co.models.zoo import HetGNNModel
+from rl4co.models.zoo import L2DModel
```
fedebotu, line 22: Good - let's make sure the baselines have their names! L2D is a very influential paper in the NCO community

rl4co/utils/ops.py
```diff
@@ -73,7 +73,8 @@ def gather_by_index(src, idx, dim=1, squeeze=True):
     expanded_shape = list(src.shape)
     expanded_shape[dim] = -1
     idx = idx.view(idx.shape + (1,) * (src.dim() - idx.dim())).expand(expanded_shape)
-    return src.gather(dim, idx).squeeze() if squeeze else src.gather(dim, idx)
+    squeeze = idx.size(dim) == 1 and squeeze
+    return src.gather(dim, idx).squeeze(dim) if squeeze else src.gather(dim, idx)
 
 
 def unbatchify_and_gather(x: Tensor, idx: Tensor, n: int):
```
fedebotu, line 80: [Minor] it might be faster if you put a `@torch.jit` decorator. Not 100% sure though

fedebotu (approved):

fedebotu:
@Junyoungpark tagging you since you know this problem very well!

Do you think we have a chance to include [ScheduleNet](https://arxiv.org/abs/2106.03051)?

rl4co/envs/scheduling/jssp/env.py
```diff
@@ -0,0 +1,95 @@
...
+from tensordict import TensorDict
+from torch._tensor import Tensor
+
+from rl4co.envs import FJSPEnv
+from rl4co.utils.ops import gather_by_index
+
+from .generator import JSSPFileGenerator, JSSPGenerator
+
+
+class JSSPEnv(FJSPEnv):
```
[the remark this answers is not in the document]
cbhua: Yesss, when see this inherit: wow nice.

rl4co/models/nn/graph/gcn.py
```diff
@@ -15,7 +15,104 @@
...
 
+EdgeIndexFnSignature = Callable[[TensorDict, int, bool], Tensor]
+
+
+def edge_idx_fn_wrapper(td: TensorDict, num_nodes: int):
+    # self-loop is added by GCNConv layer
+    return get_full_graph_edge_index(td.device, num_nodes, self_loop=False)
+
+
 class GCNEncoder(nn.Module):
```
cbhua, line 26: I like this clean refactoring, the logic is clearer. But will the `get_full_graph_edge_index()` be called at every forward step? i.e. in the previous version, if it's a fully connected connected graph, the `edge_index` will be saved as class variable, instead of regenerating every time.

cbhua (approved):
Great job!

rl4co/envs/scheduling/fjsp/env.py
```diff
@@ -285,6 +293,17 @@ def _make_step(self, td: TensorDict, selected_job, selected_machine) -> TensorDi
         td["ma_assignment"][batch_idx, selected_machine, selected_op] = 1
         # update the state of the selected machine
         td["busy_until"][batch_idx, selected_machine] = td["time"] + proc_time_of_action
+        # update adjacency matrices (remove edges)
+        td["proc_times"] = td["proc_times"].scatter(
+            2,
+            selected_op[:, None, None].expand(-1, self.num_mas, 1),
```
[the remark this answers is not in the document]
fedebotu: Is it though? I think if you know already the dimensions, `einops` is slightly slower from what I know, but take this with a grain of salt. But I agree that it's more readable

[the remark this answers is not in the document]
cbhua: Okay I did a trial and actually `einops.repeat` is way slower than `tensor.expand` 😂 (in large scale around 4x slower). Then I think it's good to keep using the `tensor.expand`.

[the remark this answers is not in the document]
LTluttmann: yeah I think they use very clever optimizations in torch.expand() 😄

rl4co/models/nn/graph/gcn.py
LTluttmann: thats true, but the result is cached so it should not be too slow. In fact, this implementation should be much faster as before (at least it was in my experiments), because it avoids the list comprehension over the batch data within the forward pass. But I agree, its still not optimal; I will revisit this in the near future

LTluttmann:
Thanks for reviewing guys. I will add a ton of changes here in a couple of minutes and I hope this PR will not get too messy. Let me know if we should go through it together.

Additional changes:
- stepwise PPO for L2D 
- Some MatNet changes to better work for JSSP / FJSP
- Running mean / variance class for reward / advantage scaling
- stepwise L2D policy
- attention based models for fjsp / jssp
- minor bugfixes and improvements here and there

fedebotu:
Wow, lots of changes here! 😁 Really curious about episodic / stepwise RL performances


Btw, feel free to merge anytime

rl4co/envs/routing/tsp/env.py
```diff
@@ -167,5 +167,59 @@ def check_solution_validity(td: TensorDict, actions: torch.Tensor):
...
     @staticmethod
-    def render(td: TensorDict, actions: torch.Tensor=None, ax = None):
+    def render(td: TensorDict, actions: torch.Tensor = None, ax=None):
         return render(td, actions, ax)
+
+
+# NOTE Experimental TSP class for stepwise PPO
+
+
+class TSPEnv4PPO(TSPEnv):
```
fedebotu, line 177: [Minor] this may be called `DenseRewardTSPEnv` or similar?

rl4co/envs/scheduling/ffsp/env.py
```diff
@@ -58,7 +57,7 @@ def __init__(
         generator_params: dict = {},
         **kwargs,
     ):
-        super().__init__(**kwargs)
+        super().__init__(check_solution=False, **kwargs)
```
fedebotu, line 60: Is this always the case (no solution check)?

fedebotu (approved):

rl4co/envs/routing/tsp/env.py
LTluttmann: Sure! Btw., Stepwise PPO for TSP indeed converges to the nearest neighbor heuristic, at least with the stepwise reward as it is defined here (the distance added by the action):

image

Do you have a preferrence as to how to call the stepwise PPO in the paper (dense, stepwise something else)? And then we should probably adjust the description about PPO in the appendix

rl4co/envs/scheduling/ffsp/env.py
LTluttmann: uhm I think this is there just bc there is no check implemented for FFSP yet haha. Let me see if I can get one implemented

fedebotu: No worries, this is not a pressing issue! Actually, it's even better to keep it to `False` during training for efficiency

fedebotu:
Great job!! This PR is truly huge

PR #180, opened 2024-05-24
[BugFix] solve memory leakage during autoregressive decoding

LTluttmann:
## Description

By not keeping only the logprobs of selected actions but all logprobs in the buffer of the decoding strategies, memory issues might occur. This is now fixed by adding an additional parameter `only_store_selected_logp` to the base decoding strategy, which defaults to True. Only exception is when we require the computation of the entropy (as in PPO). 
This fix should be backward compatible. We can still pass all logprobs to the `get_log_likelihood` function along with the actions, in which case this functions gathers the correct logprobs (like before).

## Motivation and Context

see above

## Types of changes

## Checklist

### DIFF
--- a/rl4co/models/common/constructive/base.py
+++ b/rl4co/models/common/constructive/base.py
@@ -212,6 +212,7 @@ class ConstructivePolicy(nn.Module):
             temperature=decoding_kwargs.pop("temperature", self.temperature),
             tanh_clipping=decoding_kwargs.pop("tanh_clipping", self.tanh_clipping),
             mask_logits=decoding_kwargs.pop("mask_logits", self.mask_logits),
+            only_store_selected_logp=not return_entropy,
             **decoding_kwargs,
         )
 
--- a/rl4co/models/common/constructive/base.py
+++ b/rl4co/models/common/constructive/base.py
@@ -212,7 +212,7 @@ class ConstructivePolicy(nn.Module):
             temperature=decoding_kwargs.pop("temperature", self.temperature),
             tanh_clipping=decoding_kwargs.pop("tanh_clipping", self.tanh_clipping),
             mask_logits=decoding_kwargs.pop("mask_logits", self.mask_logits),
-            only_store_selected_logp=not return_entropy,
+            store_all_logp=decoding_kwargs.pop("store_all_logp", return_entropy),
             **decoding_kwargs,
         )
 
--- a/rl4co/utils/decoding.py
+++ b/rl4co/utils/decoding.py
@@ -285,7 +285,10 @@ class DecodingStrategy(metaclass=abc.ABCMeta):
             td.set("action", action)
             td = env.step(td)["next"]
             # first logprobs is 0, so p = logprobs.exp() = 1
-            logprobs = torch.zeros_like(action, device=td.device)
+            if self.store_all_logp:
+                logprobs = torch.zeros_like(td["action_mask"])  # [B, N]
+            else:
+                logprobs = torch.zeros_like(action, device=td.device)  # [B]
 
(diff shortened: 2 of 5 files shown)


rl4co/models/common/constructive/base.py
```diff
@@ -212,6 +212,7 @@ def forward(
             temperature=decoding_kwargs.pop("temperature", self.temperature),
             tanh_clipping=decoding_kwargs.pop("tanh_clipping", self.tanh_clipping),
             mask_logits=decoding_kwargs.pop("mask_logits", self.mask_logits),
+            only_store_selected_logp=not return_entropy,
```
fedebotu, line 215: This is the only case; right? Otherwise, we could pop this from the kwargs like

```python
kwargs.pop("only_store_selected_logp", not return_entropy)
```

rl4co/utils/decoding.py
```diff
@@ -418,6 +423,8 @@ class BeamSearch(DecodingStrategy):
     name = "beam_search"
 
     def __init__(self, beam_width=None, select_best=True, **kwargs) -> None:
+        # TODO do we really need all logp in beam search?
+        kwargs["only_store_selected_logp"] = False
```
fedebotu, line 427: It looks like its needed in beam search, right? I'm no expert in this, though

fedebotu (approved):
Awesome!

I will do some small experiment in training speed and memory consumption with / without this improvement. As we noticed this can be substantial at least in large scale inference, so I'm really curious how much training speed will be impacted :)

cc @cbhua @ngastzepeda @ahottung

rl4co/models/common/constructive/base.py
LTluttmann: nice idea using `pop`, I will adapt the PR accordingly

rl4co/utils/decoding.py
LTluttmann: Currently all logps are used but I am not certain whether this is indeed necessary. I will debug through beam search after the submission

fedebotu: All right, yes this would be minor since it is not really used during training

rl4co/utils/decoding.py
```diff
@@ -280,9 +284,8 @@ def pre_decoder_hook(
 
             td.set("action", action)
             td = env.step(td)["next"]
-            logprobs = torch.zeros_like(
-                td["action_mask"], device=td.device
-            )  # first logprobs is 0, so p = logprobs.exp() = 1
+            # first logprobs is 0, so p = logprobs.exp() = 1
+            logprobs = torch.zeros_like(action, device=td.device)
```
fedebotu, line 288: Note: If we are storing all log p, I guess we should put either
` logprobs = torch.zeros_like(td["action_mask"])`, otherwise, it will be  `torch.zeros_like(action)`, right?

fedebotu: Done

fedebotu:
Results on a notebook, single 3090, training POMO with 100k samples/epoch,  bs 64, on  TSP 100 
Old / New
- Time: 3:02 / **2:58** 
- GPU load: 61-68% / **59-65%**
- VRAM: 5.5GB /  **3.4GB**

The training time and load improve slightly (some percentage points) and VRAM usage is decreased by around 40%! :rocket:  Will merge

cbhua:
Thanks @LTluttmann! Awesome work. I didn't give a review in time but based on the test results as shown by @fedebotu , I believe this is a really good updating! 🙌

rl4co/utils/decoding.py
LTluttmann: great catch!

PR #181, opened 2024-05-26
[Feat] Add MVMoE model

RoyalSkye:
## Description

This PR adds support for [`MVMoE`](https://arxiv.org/pdf/2405.01029) in the multi-task VRP setting. In specific, the token-level [input-choice](https://arxiv.org/pdf/1701.06538) gating is implemented. Some minor bugs are fixed as well, such as missing the definition of `MTVRPInitEmbedding` and `MTVRPContext`. 

* [x] An example will be given in `1-mtvrp.ipynb`.

## Motivation and Context

A step towards foundation VRP models.

## Types of changes

- [x] Example (update in the folder of examples)

## Checklist


rl4co/models/nn/graph/attnnet.py
```diff
@@ -29,19 +32,18 @@ def __init__(
         feedforward_hidden: int = 512,
         normalization: Optional[str] = "batch",
         sdpa_fn: Optional[Callable] = None,
+        moe_kwargs: Optional[dict] = None,
```
cbhua, line 35: Same as the pointer attention, we may want a separate class for MoE's MHA.

rl4co/models/nn/env_embeddings/init.py
```diff
@@ -444,3 +445,25 @@ def _stepwise_operations_embed(self, td: TensorDict):
...
+        # [!] convert [0, inf] -> [0, 0] if a problem does not include the time window constraint, do not modify in-place
+        time_windows = torch.where(torch.isinf(time_windows), torch.tensor(0.), time_windows)
+        # embeddings
+        depot_embedding = self.init_embed_depot(depot)
+        node_embeddings = self.init_embed(
+            torch.cat(
+                (cities, demand_linehaul[..., None], demand_backhaul[..., None], time_windows, service_time[..., None]), -1
+            )
+        )
+        return torch.cat((depot_embedding, node_embeddings), -2)
```
cbhua, line 469: Do we need to record the moe loss here?

cbhua (commented):
Great job! Thanks for your work! Left some discussion points~ 😁

rl4co/models/nn/env_embeddings/init.py
fedebotu: > Do we need to record the moe loss here?

If we use the raw feature MoE yes, but @RoyalSkye showed this does not work too well, so we might skip I guess?

rl4co/models/nn/env_embeddings/init.py
```diff
@@ -444,3 +445,25 @@ def _stepwise_operations_embed(self, td: TensorDict):
...
+        # node_dim = 7: x, y, demand_linehaul, demand_backhaul, tw start, tw end, service time
+        super(MTVRPInitEmbedding, self).__init__(embed_dim, linear_bias, node_dim)
+
+    def forward(self, td):
+        depot, cities = td["locs"][:, :1, :], td["locs"][:, 1:, :]
+        demand_linehaul, demand_backhaul = td["demand_linehaul"][..., 1:], td["demand_backhaul"][..., 1:]
+        service_time = td["service_time"][..., 1:]
+        time_windows = td["time_windows"][..., 1:, :]
+        # [!] convert [0, inf] -> [0, 0] if a problem does not include the time window constraint, do not modify in-place
+        time_windows = torch.where(torch.isinf(time_windows), torch.tensor(0.), time_windows)
```
fedebotu, line 461: The same can be achieved slightly more efficiently with:

```python
 time_windows = torch.nan_to_num(time_windows,  posinf=0.0)
```

rl4co/models/zoo/mvmoe/model.py
```diff
@@ -0,0 +1,79 @@
...
+                "normalization": "instance",
+                "use_graph_context": False,
+                "moe_kwargs": moe_kwargs,
+            }
+            policy_kwargs.update(policy_kwargs_)
+            policy = AttentionModelPolicy(env_name=env.name, **policy_kwargs)
+
+        # Initialize with the shared baseline
+        super(MVMoE_POMO, self).__init__(env, policy, policy_kwargs, baseline, num_augment, augment_fn,
+                                         first_aug_identity, feats, num_starts, **kwargs)
```
fedebotu, line 49: It seems that the MoE loss is not included, right?
(I sent you on Slack a reference)

rl4co/models/nn/attention.py
```diff
@@ -145,6 +146,7 @@ class PointerAttention(nn.Module):
         linear_bias: whether to use bias in linear projection
         check_nan: whether to check for NaNs in logits
         sdpa_fn: scaled dot product attention function (SDPA) implementation
+        moe_kwargs: Keyword arguments for MoE
     """
 
     def __init__(
```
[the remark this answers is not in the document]
fedebotu: Mmmh... In principle, it might be better to have all the contributions of a model in its own repo. But given that such contribution could improve other models (say: if you wanted, you may try applying MoE to scheduling) and **if it does not change the base performance** (which seems to be the case here?), then I think it can be good. What do you think?

Asking also master @LTluttmann

rl4co/models/zoo/mvmoe/model.py
RoyalSkye: Yes, we recently found the MoE loss may conflict with the reinforcement loss during optimization, and thus I ignored it. Should we add it?

rl4co/models/nn/env_embeddings/init.py
RoyalSkye: Please see the below comments.

rl4co/models/nn/attention.py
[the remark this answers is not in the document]
RoyalSkye: Yes, maybe I can create a new `class PointerAttention_MoE(PointerAttention)`. Let me know your final decision.

rl4co/models/zoo/mvmoe/model.py
cbhua: If the MoE loss doesn't contribute much to the performance I think we could just skip it.

rl4co/models/nn/attention.py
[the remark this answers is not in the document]
LTluttmann: I would also go for a separate `PointerAttnMoE(PointerAttention)` class for MoE, which would only need to overwrite the `project_out`. I guess we have the best of both worlds then - most of the `PointerAttention` code is reused while we separate the logic in a clear manner.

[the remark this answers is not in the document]
RoyalSkye: okay, I will finish this part in the next commit!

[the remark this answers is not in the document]
RoyalSkye: Hi all, I have finished this part. As for the MHA in `attnnet.py`, I think the current code is neat, so I keep it as is. In my opinion, MoE is a general idea, which may benefit various CO tasks. So implementing it in the underlying code may be a better choice..? What do you think? Kindly let me know if any problem exists.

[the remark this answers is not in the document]
fedebotu: On that I agree, it doesn't impact performance in MHA in `attnnet.py` correct? If so, I'm good keeping it, so people may explore it!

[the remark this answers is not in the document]
RoyalSkye: Yes, the default setting is `moe_kwargs=None`. In this case, it's equivalent to the original MHA, and thus the performance will not be affected.

rl4co/models/zoo/am/decoder.py
```diff
@@ -108,7 +108,7 @@ def __init__(
 
         if pointer is None:
             # MHA with Pointer mechanism (https://arxiv.org/abs/1506.03134)
-            pointer = PointerAttention(
```
fedebotu, line 109: So we are defaulting to `PointerAttnMoE` here for the base AM class? I think this might be confusing for some who just want the default model (without MoE). I would rather keep the previous version in this case (`PointerAttention` but with MoE kwargs)

Or, I would suggest to do this:
```python
pointer_attn_class = PointerAttention if moe_kwargs is None else PointerAttnMoE
pointer = pointer_attn_class(embed_dim ...)
```

rl4co/models/nn/attention.py
```diff
@@ -220,6 +211,53 @@ def _inner_mha(self, query, key, value, attn_mask):
...
+        if self.moe_kwargs is not None:
+            self.project_out = None
+            self.project_out_moe = MoE(embed_dim, embed_dim, num_neurons=[], out_bias=out_bias, **moe_kwargs)
+            if self.moe_kwargs["light_version"]:
+                self.dense_or_moe = nn.Linear(embed_dim, 2, bias=False)
+                self.project_out = nn.Linear(embed_dim, embed_dim, bias=out_bias)
+
     def _project_out(self, out):
         """Implementation of Hierarchical Gating based on Zhou et al. (2024) <https://arxiv.org/abs/2405.01029>."""
         if self.moe_kwargs is not None:
```
fedebotu, line 263: (read comment below) if `PointerAttnMoe` by default has `MoE`, then we could also remove this "if, else" statement

rl4co/models/zoo/am/decoder.py
RoyalSkye: The second one looks good! Please see the latest commit.

fedebotu (approved):
:rocket: 

@cbhua @LTluttmann any of you who is the last one to approve can merge!

cbhua (approved):
Nice work, let go~

RoyalSkye:
I have added an example to `1-mtvrp.ipynb`. It seems that the notebook has some issues on my laptop, so I just uploaded a raw version without output 🚀.

fedebotu:
> It seems that the notebook has some issues on my laptop, so I just uploaded a raw version without output 

Are you using an M-series Macbook? We had some problems only with MTVRP and only with Apple MPS accelerators :rofl:

RoyalSkye:
Thank Chuanbo. Looks great! I have nothing to add to this PR.

cbhua:
Great job! After @LTluttmann having a look let's launch! 🚀

LTluttmann:
awesome work! Clean and well documented code, was a pleasure to read it haha! Merging now

PR #183, opened 2024-05-27
Add meta learning framework

jieyibi:
## Description

Add a new training framework based on meta learning. Details refer to [Zhou et al. 2023](https://arxiv.org/abs/2305.19587). 

## Motivation and Context

To address the generalization issue. 

## Types of changes

- [x] Example (update in the folder of examples)

## Checklist

### DIFF
--- a/examples/2d-meta_train.py
+++ b/examples/2d-meta_train.py
@@ -1,14 +1,11 @@
-import pytz
-import torch
-
-from datetime import datetime
 from lightning.pytorch.callbacks import ModelCheckpoint, RichModelSummary
 from lightning.pytorch.loggers import WandbLogger
 
 from rl4co.envs import CVRPEnv
 from rl4co.models.zoo.am import AttentionModelPolicy
 from rl4co.models.zoo.pomo import POMO
-from rl4co.utils.meta_trainer import RL4COMetaTrainer, MetaModelCallback
+from rl4co.utils.trainer import RL4COTrainer
+from rl4co.utils.meta_trainer import ReptileCallback
 
 def main():
     # Set device
@@ -34,7 +31,6 @@ def main():
                  train_data_size=64 * 50,  # each epoch
                  val_data_size=0,
                  optimizer_kwargs={"lr": 1e-4, "weight_decay": 1e-6},
-                 # for the task scheduler of size setting, where sch_epoch = 0.9 * epochs
                  )
 
     # Example callbacks
@@ -47,10 +43,10 @@ def main():
         mode="max",  # maximize validation reward
     )
     rich_model_summary = RichModelSummary(max_depth=3)  # model summary callback
+
     # Meta callbacks
-    meta_callback = MetaModelCallback(
+    meta_callback = ReptileCallback(
         meta_params={
-            'meta_method': 'reptile',  # choose from ['maml', 'fomaml', 'maml_fomaml', 'reptile']
             'data_type': 'size',  # choose from ["size", "distribution", "size_distribution"]
             'sch_bar': 0.9,  # for the task scheduler of size setting, where sch_epoch = sch_bar * epochs
             'B': 1,  # the number of tasks in a mini-batch
@@ -64,13 +60,12 @@ def main():
     callbacks = [meta_callback, checkpoint_callback, rich_model_summary]
 
     # Logger
-    process_start_time = datetime.now(pytz.timezone("Asia/Singapore"))
-    logger = WandbLogger(project="rl4co", name=f"{env.name}_{process_start_time.strftime('%Y%m%d_%H%M%S')}")
+    logger = WandbLogger(project="rl4co", name=f"{env.name}_pomo_reptile")
     # logger = None # uncomment this line if you don't want logging
 
     # Adjust your trainer to the number of epochs you want to run
-    trainer = RL4COMetaTrainer(
-        max_epochs=20000,  # (the number of meta-model updates) * (the number of tasks in a mini-batch)
+    trainer = RL4COTrainer(
+        max_epochs=20000,  # (the number of meta_model updates) * (the number of tasks in a mini-batch)
         callbacks=callbacks,
         accelerator="gpu",
         devices=[device_id],
--- a/examples/2d-meta_train.py
+++ b/examples/2d-meta_train.py
@@ -1,3 +1,6 @@
+import sys
+sys.path.append("/home/jieyi/rl4co")
+
 from lightning.pytorch.callbacks import ModelCheckpoint, RichModelSummary
 from lightning.pytorch.loggers import WandbLogger
 
@@ -9,7 +12,7 @@ from rl4co.utils.meta_trainer import ReptileCallback
 
 def main():
     # Set device
-    device_id = 0
+    device_id = 1
 
     # RL4CO env based on TorchRL
     env = CVRPEnv(generator_params={'num_loc': 50})
@@ -46,13 +49,13 @@ def main():
 
     # Meta callbacks
     meta_callback = ReptileCallback(
-        num_tasks = 1,  # the number of tasks in a mini-batch
-        alpha = 0.99,  # params for the outer-loop optimization of reptile
-        alpha_decay = 0.999,  # params for the outer-loop optimization of reptile
-        min_size = 20,  # minimum of sampled size in meta tasks
-        max_size= 150,  # maximum of sampled size in meta tasks
-        data_type="size",  # choose from ["size", "distribution", "size_distribution"]
-        sch_bar=0.9,  # for the task scheduler of size setting, where sch_epoch = sch_bar * epochs
+        num_tasks = 1,  # the number of tasks in a mini-batch, i.e. `B` in the original paper
+        alpha = 0.99,  # initial weight of the task model for the outer-loop optimization of reptile
+        alpha_decay = 0.999,  # weight decay of the task model for the outer-loop optimization of reptile
+        min_size = 20,  # minimum of sampled size in meta tasks (only supported in cross-size generalization)
+        max_size= 150,  # maximum of sampled size in meta tasks (only supported in cross-size generalization)
+        data_type="size_distribution",  # choose from ["size", "distribution", "size_distribution"]
+        sch_bar=0.9,  # for the task scheduler of size setting, where lr_decay_epoch = sch_bar * epochs, i.e. after this epoch, learning rate will decay with a weight 0.1
         print_log=True # whether to print the sampled tasks in each meta iteration
     )
     callbacks = [meta_callback, checkpoint_callback, rich_model_summary]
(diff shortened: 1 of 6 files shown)


rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,245 @@
...
+        # Load new training task (Update the environment)
+        task_params, task_w = self.selected_tasks[task_idx], self.w[task_idx].item()
+        task_capacity = math.ceil(30 + task_params[0] / 5) if task_params[0] >= 20 else 20
+        pl_module.env.generator.num_loc = task_params[0]
+        pl_module.env.generator.capacity = task_capacity
+
+    def _alpha_scheduler(self):
+        self.alpha = max(self.alpha * self.alpha_decay, 0.0001)
+
+class RL4COMetaTrainer(Trainer):
```
fedebotu, line 105: Are there any differences compared to the [`RL4COTrainer`](https://github.com/ai4co/rl4co/blob/03b31da68146915860e44fb5584a20596159ff1f/rl4co/utils/trainer.py#L17)? I could not find any at a first glance. I guess the only difference is that we pass the `MetaModelCallback`?

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,245 @@
...
+from lightning.pytorch.core.datamodule import LightningDataModule
+from lightning.pytorch.loggers import Logger
+from lightning.pytorch.strategies import DDPStrategy, Strategy
+from lightning.pytorch.utilities.types import EVAL_DATALOADERS, TRAIN_DATALOADERS
+from rl4co import utils
+import random
+log = utils.get_pylogger(__name__)
+
+
+class MetaModelCallback(Callback):
```
fedebotu, line 21: I haven't thought of Meta-learning as Lightning Callbacks, but it looks neat! :D

fedebotu: Tagging @Junyoungpark for the good ol' Reptile

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,245 @@
...
+    def __init__(self, meta_params, print_log=True):
+        super().__init__()
+        self.meta_params = meta_params
+        assert meta_params["meta_method"] == 'reptile', NotImplementedError
```
fedebotu, line 25: Another general comment: it seems that this callback is only for Reptile, so we should consider calling it `ReptileCallback`. I think it would be cool to have these as say a `meta_learning/` folder

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,245 @@
...
+        self.alpha = self.meta_params["alpha"]
+        self.alpha_decay = self.meta_params["alpha_decay"]
+        self.sch_bar = self.meta_params["sch_bar"]
+        self.task_set = [(n,) for n in range(self.meta_params["min_size"], self.meta_params["max_size"] + 1)]
+
+        # Sample a batch of tasks
+        self._sample_task()
+        self.selected_tasks[0] = (pl_module.env.generator.num_loc, )
+
+    def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
```
fedebotu, line 41: [Minor] type hints should be classes, not strings. For example `Trainer` and `LightningModule` (Maybe even better: `RL4COTrainer` and `RL4COLitModule` since everything is inherited from that)

fedebotu (commented):
Great! :rocket: 

Left some comments here and there~

examples/2d-meta_train.py
```diff
@@ -0,0 +1,87 @@
+import pytz
```
cbhua, line 1: [Minor] Nice tool but `pytz` is not included in RL4CO's dependence packages. Maybe better to have a package check here: 

```python
try:
    import pytz
except ImportError:
    # raise a warning and use python default timeit
    pass
```

rl4co/utils/meta_trainer.py
jieyibi: Yes. `ReptileCallback` is better. Where should this new `meta_learning/` folder be located? In `utils/`?

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,245 @@
...
+import random
+log = utils.get_pylogger(__name__)
+
+
+class MetaModelCallback(Callback):
+    def __init__(self, meta_params, print_log=True):
+        super().__init__()
+        self.meta_params = meta_params
+        assert meta_params["meta_method"] == 'reptile', NotImplementedError
+        assert meta_params["data_type"] == 'size', NotImplementedError
```
cbhua, line 26: It seems that this parameter `data_type` is not called anywhere? 🤔

rl4co/utils/meta_trainer.py
fedebotu: I''d say under `models/rl` (which includes `LightningModules`), since `meta_learning` is a way to optimize a policy

cbhua (approved):
Great job! 🚀 Really nice to have meta learning supporting 😁

rl4co/utils/meta_trainer.py
jieyibi: Hi. I am wondering refactor the `ReptileCallback' inherited from the `REINFORCE` and the `RL4COLitModule`. But in this case, maybe I need to add new `meta_model.py` inherited the `model.py` under the `zoo/pomo/` folder or the `zoo/am/` folder to call the REPTILE outside. It is a little bit redundant... Maybe a Lightning Callback is more generic. We could apply it to every model and every policy.

examples/2d-meta_train.py
jieyibi: Yeah, maybe it could be removed.

fedebotu (commented):
Great! 
Also:

1. Could you reproduce the learning curves of the original model?
2. Could you make a [simple test](https://github.com/ai4co/rl4co/blob/6d1df3152716bfa199356cd4abab6843fc7244d9/tests/test_training.py#L82) for your model so that it can be automatically checked?

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,100 @@
...
+from rl4co import utils
+import random
+log = utils.get_pylogger(__name__)
+
+
+class ReptileCallback(Callback):
+
+    # Meta training framework for addressing the generalization issue
+    # Based on Zhou et al. (2023): https://arxiv.org/abs/2305.19587
+    def __init__(self, meta_params, print_log=True):
```
fedebotu, line 17: [Minor] I think it would be more clear to list the Hyperparameters.

For  example

```python
def __init__(self, 
alpha=...,
alpha_decay=...
... 
print_log=True):
```

which is generally easier to maintain and to document.

fedebotu:
It seems there was a mistake [here](https://github.com/ai4co/rl4co/actions/runs/9265448786/job/25493241566?pr=183#:~:text=FAILED%20tests/test_training,389) - I noticed that the "tasks" are a bit hardcoded, i.e. the "size" and "capacity" (I guess for TSP and CVRP, right?)

---

Speaking of testing: you can make sure things work on your device before committing by using `pytest tests`

rl4co/utils/meta_trainer.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self,
+                 num_tasks,
+                 alpha,
+                 alpha_decay,
+                 min_size,
+                 max_size,
+                 sch_bar = 0.9,
+                 data_type = "size",
+                 print_log=True):
+        super().__init__()
```
cbhua, line 26: [Documentation] It's recommended to have a doc with parameters, possibly including the data type, constraints, hints, etc. Better for us "non-experts" to understand 😆

```python
class ReptileCallback(Callback):
    """Meta training framework for addressing the generalization issue
    Based on Zhou et al. (2023): https://arxiv.org/abs/2305.19587

    Args:
        - num_tasks: number of task types, i.e. `B` in the original paper
        - alpha: ...
        - ...
    """

    def __init__(
        self,
        num_tasks: int,
        alpha: float,
        alpha_decay: float,
        min_size: int,
        max_size: int,
        sch_bar: float = 0.9,
        data_type: str = "size",
        print_log: bool = True,
    ):
        super().__init__()
```

jieyibi: Hi Chuanbo. I have added the documentation as you recommended, along with the generation code for some distributions defined in the generalization-related works. Now the meta learning framework is supported for cross-distribution generalization.

rl4co/envs/common/distribution_utils.py
```diff
@@ -0,0 +1,184 @@
+import torch
+
+class Cluster():
```
fedebotu, line 3: I like this!  @cbhua I think we should train some model with say TSP50 / CVRP50 with a mixed distribution and test its generalization performance

Minor comment: Shouldn't this be a subclass of `torch.distributions.distribution.Distribution`?

examples/2d-meta_train.py
```diff
@@ -1,3 +1,6 @@
+import sys
+sys.path.append("/home/jieyi/rl4co")
```
fedebotu: [Minor] remember to remove this - if you install `rl4co` locally in editable mode with `pip install -e .`, importing a path should not be needed

rl4co/envs/common/distribution_utils.py
cbhua: Yes actually this class is what we are missing in the distribution! This could be used by various environments' generator. 

About the experiment, we want to test the distribution generalization ability right?

examples/2d-meta_train.py
jieyibi: Yes, already removed in the latest commit:)

fedebotu:
Routine check, how is progress going? I think the multi-distribution generators in particular should be included, since they are part of `MDPOMO`, right?

jieyibi:
Maybe we could include the MDPOMO one first. I will commit a new PR to support the MDPOMO.

jieyibi:
The updated commit supports training on multiple mixed distributions by changing the argument `loc_distribution` to "mix_distribution"  in the `generator_params` of the environment, e.g. 

`env = CVRPEnv(generator_params={'loc_distribution': "mix_distribution"})`

The rest of the arguments remain the same.

Note that this mixed distribution setting follows the # setting in Bi et al., 2022.

fedebotu:
Great! Should the code be merged?

jieyibi:
Yes. I think it is ready to merge. The newly reproduced performance (by changing some key parameters) seems similar to that in the literature.

fedebotu (approved):

fedebotu:
Great!

How about the generalization experiments, i.e., MDPOMO? Do you have `yaml` configuration for them like [this](https://github.com/ai4co/rl4co/blob/main/configs/experiment/routing/am.yaml), or a training script?

jieyibi:
Yeah, already added to the main branch:)

fedebotu:
Awesome! Then we can go ahead and merge :)

PR #191, opened 2024-06-10
[Feat] FLP and MCP and enviroments

bokveizen:
## Description

Added environments for the Facility Location Problem (FLP) and the Maximum Coverage Problem (MCP).

## Motivation and Context

Added various problem environments other than routing problems.

## Types of changes

## Checklist


rl4co/utils/ops.py
```diff
@@ -147,7 +147,7 @@ def select_start_nodes(td, env, num_starts):
         num_starts: Number of nodes to select. This may be passed when calling the policy directly. See :class:`rl4co.models.AutoregressiveDecoder`
     """
     num_loc = env.generator.num_loc if hasattr(env.generator, "num_loc") else 0xFFFFFFFF
-    if env.name in ["tsp", "atsp"]:
+    if env.name in ["tsp", "atsp", "flp", "mcp"]:
```
fedebotu, line 150: We may use the `env.select_start_nodes` as well, but this is also OK as a fallback method

rl4co/envs/graph/mcp/render.py
```diff
@@ -0,0 +1,14 @@
...
+import torch
+
+from rl4co.utils.ops import gather_by_index
+from rl4co.utils.pylogger import get_pylogger
+
+log = get_pylogger(__name__)
+
+
+def render(td, actions=None, ax=None):
+    # TODO: better rendering, e.g., visualization
```
fedebotu, line 12: Something missing here I guess? If not implemented, we can simply do `raise NotImplementedError`

rl4co/envs/graph/flp/render.py
```diff
@@ -0,0 +1,14 @@
...
+import torch
+
+from rl4co.utils.ops import gather_by_index
+from rl4co.utils.pylogger import get_pylogger
+
+log = get_pylogger(__name__)
+
+
+def render(td, actions=None, ax=None):
+    # TODO: better rendering, e.g., visualization
```
fedebotu, line 12: Same comment as MCP

rl4co/envs/graph/__init__.py
```diff
@@ -0,0 +1,3 @@
+from rl4co.envs.graph.mcp.env import MCPEnv
+from rl4co.envs.graph.flp.env import FLPEnv
```
fedebotu, line 2: [Minor] We should add these envs to the `rl4co.envs.__init__.py` so they are importable from there too

fedebotu (commented):
Great! Mostly minor comments have been added.
Once we have some model, we should also add some test under `tests` (progress to see what is being tested and what not can be checked under [Codecov](https://app.codecov.io/github/ai4co/rl4co))

rl4co/utils/ops.py
bokveizen: I see! I do think we may need some different ways of "selected_start_nodes" for non-routing problems since choosing all distinct starting nodes might not be reasonable for some envs.

rl4co/envs/graph/__init__.py
bokveizen: Added! Thanks!

rl4co/utils/ops.py
fedebotu: Yes, exactly. If the function is provided in the env, by default it will be provided

rl4co/envs/graph/mcp/render.py
bokveizen: Thanks. I simply removed it for now.

rl4co/utils/ops.py
bokveizen: Where can I see an example of "env.select_start_nodes being provided"?

fedebotu: For example in PDP environment [here](https://github.com/ai4co/rl4co/blob/2e333356a2d20e9661b9b0793d7b5cc4b9fd53f0/rl4co/envs/routing/pdp/env.py#L222) only pickup nodes can be selected as first node.



By default, [this](https://github.com/ai4co/rl4co/blob/2e333356a2d20e9661b9b0793d7b5cc4b9fd53f0/rl4co/utils/decoding.py#L291) is called (i.e. it calls the class of RL4COEnvBase [here](https://github.com/ai4co/rl4co/blob/2e333356a2d20e9661b9b0793d7b5cc4b9fd53f0/rl4co/envs/common/base.py#L206) which is provided by the function in `utils`. If `select_start_nodes` is overwritten, then it will be taken from that environment: this is preferable and more modular, and eventually we would like to change to this behavior

fedebotu:
It seems that most comments were resolved, should we go ahead and merge?

bokveizen:
> It seems that most comments were resolved, should we go ahead and merge?

I think the core functions are okay now! I may add some additional functions, e.g., render later :D

fedebotu:
Yep, we can do rendering afterwards. I think the only major thing missing for the `0.5.0` release is an overhaul of documentation :+1:

PR #204, opened 2024-08-02
[BugFix] python version

ShuN6211:
Enable to run "poetry install".

## Description

Update pyproject.toml to be enable to run `poetry install` and install rl4co project by poetry tool.
At current project, running `poetry install` may causes following errors.

```
The current project's Python requirement (>=3.8) is not compatible with some of the required packages Python requirement:
  - pyvrp requires Python <4.0,>=3.9, so it will not be satisfied for Python >=3.8,<3.9 || >=4.0
  - pyvrp requires Python <4.0,>=3.9, so it will not be satisfied for Python >=3.8,<3.9 || >=4.0
  - pyvrp requires Python <4.0,>=3.9, so it will not be satisfied for Python >=3.8,<3.9 || >=4.0
Because no versions of pyvrp match >0.8.2,<0.9.0 || >0.9.0,<0.9.1 || >0.9.1
 and pyvrp (0.8.2) requires Python <4.0,>=3.9, pyvrp is forbidden.
And because pyvrp (0.9.0) requires Python <4.0,>=3.9, pyvrp is forbidden.
So, because pyvrp (0.9.1) requires Python <4.0,>=3.9
 and rl4co depends on pyvrp (>=0.8.2), version solving failed.
  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    For pyvrp, a possible solution would be to set the `python` property to ">=3.9,<4.0"
    For pyvrp, a possible solution would be to set the `python` property to ">=3.9,<4.0"
    For pyvrp, a possible solution would be to set the `python` property to ">=3.9,<4.0"

```
## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ packages = [{ include = "rl4co" }]
 classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
@@ -63,7 +62,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 # Required dependencies
-python = ">=3.8"
+python = ">=3.9"
 einops = "*"
 hydra-core = "*"
 hydra-colorlog = "*"
@@ -87,7 +86,7 @@ pytest-cov = { version = "*", optional = true }
 torch_geometric = { version = "*", optional = true }
 # Routing
 numba = { version = ">=0.58.1", optional = true }
-pyvrp = { version = ">=0.8.2", optional = true }
+pyvrp = { version = ">=0.8.2", optional = true, python = "<4.0" }
 # Docs
 mkdocs = { version = "*", optional = true }
 mkdocs-material = { version = "*", optional = true }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ packages = [{ include = "rl4co" }]
 classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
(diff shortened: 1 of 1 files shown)


fedebotu:
Hi @ShuN6211 , thanks for reporting this!

How about this instead:

```
pyvrp = { version = ">=0.8.2", optional = true, python = ">=3.9,<4.0" }
```

in the dependencies. This is because `pyvrp` is actually optionally and we would like not to constrain `rl4co` too much based on such dependencies. I tried `poetry install` and it works

ShuN6211:
Hi @fedebotu , thanks for commenting!!
It' good to specify python version not in dependency but in optional scope. 😃
However, only changing a line regarding "pyvrp" still causes error in my environment. 
It seems numba(optional) and lightning requires numpy and numpy requires `python = ">=3.9"`
I tried following chunk and it works.
```
[tool.poetry.dependencies]
# Required dependencies
python = ">=3.9"
# Routing
pyvrp = { version = ">=0.8.2", optional = true, python = "<4.0" }

```
Procedure
- Create some venv. ( I use miniconda to create venv)
- Activate venv
- `python -m pip install poetry`
- `rm poetry.lock`
- `poetry install`

Side notes
- machine: M1 macbook Pro 2021, macOS Ventura13.4.1
-  python version: 3.10
- poetry version:  1.5.0

fedebotu:
I see! This is better, in my opinion, since we may want to restrict a lower version of Python rather than the upper one (Python 3.8 is pretty old anyway). To be honest, though, I'm not a fan of restricting the main Python version because of some optional dependency, so if there was another way, I'd rather go for that :) 

Thoughts @cbhua @LTluttmann?

ShuN6211:
Hi @fedebotu , thanks for sharing the dropping support for python 3.8 in newest PyTorch Lightning!!
I added revert commit and new one that reflect conversation between us. If you don't prefer revert, I will reopen PR.

cbhua:
@ShuN6211 Thanks for reporting the bug about the Python version!

Considering the status of Python versions from the [official release cycle](https://devguide.python.org/versions/) that Python 3.8's end-of-life day is `2024-10`, I agree that let's stop the support for Python 3.8.

cbhua (approved):

PR #235, opened 2024-12-24
[Fix] fix the position of row and col embedding

Leaveson:
## Description

This PR addresses an issue in the MatNet implementation where the positions of the row embedding and column embedding were swapped.

## Motivation and Context

It is a significant model for handling matrices in NCO For more details, please refer to the ([original paper](https://arxiv.org/pdf/2106.11113))

## Types of changes

## Checklist


fedebotu:
Thanks for the XMas eve commit 🎄 
We discussed today with @Leaveson, and it seems good to go! 
Tagging @LTluttmann for a double check before merging since it seems we did not notice this in scheduling problems - perhaps due to the fact that the direction of row/col did not matter in those cases 👀

Issue #255, opened 2025-03-06
[BUG] MCP missing embeddings and configs


tracerr1:
The mcp.yaml is missing, so the experiments related to graph problems cannot be run directly.

fedebotu:
Thanks for your report. @bokveizen could you take a look at this?

bokveizen:
Hi, please check the pull request https://github.com/ai4co/rl4co/pull/256

tracerr1:
Thanks a lot, but MCPInitEmbedding and MCPContext are still missing.

fedebotu:
Closing as completed. Feel free to reopen if the solution does not work!

PR #257, opened 2025-03-20
[Additional features for #255] Added embedding functions for MCP

bokveizen:
Additional features for https://github.com/ai4co/rl4co/issues/255#issuecomment-2738876078
Added embedding functions for MCP.

### DIFF
--- a/rl4co/models/nn/env_embeddings/context.py
+++ b/rl4co/models/nn/env_embeddings/context.py
@@ -3,7 +3,7 @@ import torch.nn as nn
 
 from tensordict import TensorDict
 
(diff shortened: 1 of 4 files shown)


fedebotu:
LGTM :+1: 

I see that CodeCov is complaining a little. Could you add a test [here](https://github.com/ai4co/rl4co/blob/main/tests/test_policy.py)?

bokveizen:
> LGTM 👍
> 
> I see that CodeCov is complaining a little. Could you add a test [here](https://github.com/ai4co/rl4co/blob/main/tests/test_policy.py)?

Added. Seemingly okay now.

Issue #261, opened 2025-04-09
[BUG] entire dataset object being saved to hparams log


ngastzepeda:
## Describe the bug

When passing a dataset object to EAS instead of the filename, the dataset is being written to the hparams log file, leading to a hangup. 

## To Reproduce

Initializa a dataset object using `dataset = env.dataset()` and pass that to the EAS constructor (or any of the child classes).

There is no useful stacktrace, which is why the issue took longer to identify. Once the dataset object (which can be very large) is started to be written to disk, there is no logger output, which makes it seem there is nothing happening, when instead the user would expect training to start.

## System info

rl4co version 0.5.1

## Reason and Possible fixes

Fix: pass `ignore=["env", "policy", "dataset"]` to `self.save_hyperparameters()` in the constructors of the respective classes, namely `EAS` and its parent classes (`TransductiveModel` and `RL4COLitModule`).

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have provided a minimal working example to reproduce the bug (**required**)

PR #262, opened 2025-04-09
Set env, policy and dataset to be ignored in hparams for logging

ngastzepeda:
Closes issue #261 .


fedebotu (approved):
Great! As we discussed, this should do the trick :rocket:

PR #263, opened 2025-04-09
Ensure baseline for SymNCO is always "symnco" when loading from checkpoint

### LINKED ISSUE #254: AssertionError: SymNCO only supports custom-symnco baseline
I am getting the following error when trying to load the checkpoint for the SymNCO codebase. Any idea how to fix this issue? 

ERROR:
```
Traceback (most recent call last):
  File "/path//inference.py", line 128, in <module>
    solve_and_report(args.tsp_file, args.model_ckpt, args.model_name, args.num_starts)
  File "/path//inference.py", line 99, in solve_and_report
    policy = load_model(model_name, checkpoint_path, env, device)
  File "/path//inference.py", line 63, in load_model
    checkpoint = model.load_from_checkpoint(checkpoint_path, strict=False)
  File "/path//ENV-SYMNCO/lib/python3.10/site-packages/rl4co/models/rl/reinforce/reinforce.py", line 189, in load_from_checkpoint
    loaded = _load_from_checkpoint(
  File "/path//ENV-SYMNCO/lib/python3.10/site-packages/lightning/pytorch/core/saving.py", line 91, in _load_from_checkpoint
    model = _load_state(cls, checkpoint, strict=strict, **kwargs)
  File "/path//ENV-SYMNCO/lib/python3.10/site-packages/lightning/pytorch/core/saving.py", line 165, in _load_state
    obj = instantiator(cls, _cls_kwargs) if instantiator else cls(**_cls_kwargs)
  File "/path//ENV-SYMNCO/lib/python3.10/site-packages/rl4co/models/zoo/symnco/model.py", line 56, in __init__
    assert baseline == "symnco", "SymNCO only supports custom-symnco baseline"
AssertionError: SymNCO only supports custom-symnco baseline

```

```
model = SymNCO(
      env,
      policy=None,
      num_augment=2,
      num_starts=20,
      baseline='symnco').to(device)

checkpoint = model.load_from_checkpoint(checkpoint_path, strict=False)
```

I also tried loading like this, but still getting the same error.

`SymNCO.load_from_checkpoint(checkpoint_path, strict=False, load_baseline=False)`

ngastzepeda:
Closes #254.

Further, the order of dimensions `(n_start, n_aug)` in `unbatchify()` in the `shared_step()` function was fixed to `(n_aug, n_start)` to be consistent with other models in the framework. Dimensions should be `[B, A, S]`, where `B` is the batch size, `A` the number of augmentations, and `S` the number of starts.

### DIFF
--- a/rl4co/models/zoo/symnco/model.py
+++ b/rl4co/models/zoo/symnco/model.py
@@ -140,3 +140,27 @@ class SymNCO(REINFORCE):
 
         metrics = self.log_metrics(out, phase, dataloader_idx=dataloader_idx)
         return {"loss": out.get("loss", None), **metrics}
+
+    @classmethod
+    def load_from_checkpoint(
+        cls,
+        checkpoint_path,
+        map_location=None,
+        hparams_file=None,
+        strict=False,
+        load_baseline=True,
+        **kwargs,
+    ):
+        if kwargs.pop("baseline", "symnco") != "symnco":
+            log.warning(
+                "SymNCO only supports custom-symnco baseline. Setting to 'symnco'."
+            )
+        kwargs["baseline"] = "symnco"
+        return super().load_from_checkpoint(
+            checkpoint_path,
+            map_location,
+            hparams_file,
+            strict,
+            load_baseline,
+            **kwargs,
+        )
--- a/rl4co/models/zoo/symnco/model.py
+++ b/rl4co/models/zoo/symnco/model.py
@@ -87,12 +87,13 @@ class SymNCO(REINFORCE):
         out = self.policy(td, self.env, phase=phase, num_starts=n_start)
 
         # Unbatchify reward to [batch_size, n_start, n_aug].
-        reward = unbatchify(out["reward"], (n_start, n_aug))
+        unbatch_dims = (n_aug, n_start)
+        reward = unbatchify(out["reward"], unbatch_dims)
 
         # Main training loss
         if phase == "train":
             # [batch_size, n_start, n_aug]
-            ll = unbatchify(out["log_likelihood"], (n_start, n_aug))
+            ll = unbatchify(out["log_likelihood"], unbatch_dims)
 
             # Calculate losses: problem symmetricity, solution symmetricity, invariance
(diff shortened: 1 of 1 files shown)


rl4co/models/zoo/symnco/model.py
```diff
@@ -86,13 +86,14 @@ def shared_step(
         # Evaluate policy
         out = self.policy(td, self.env, phase=phase, num_starts=n_start)
 
-        # Unbatchify reward to [batch_size, n_start, n_aug].
-        reward = unbatchify(out["reward"], (n_start, n_aug))
+        # Unbatchify reward to [batch_size, n_aug, n_start].
+        unbatch_dims = (n_aug, n_start)
```
fedebotu, line 90: Much better!

rl4co/models/zoo/symnco/model.py
```diff
@@ -140,3 +141,27 @@ def shared_step(
 
         metrics = self.log_metrics(out, phase, dataloader_idx=dataloader_idx)
         return {"loss": out.get("loss", None), **metrics}
+
+    @classmethod
+    def load_from_checkpoint(
+        cls,
```
fedebotu, line 147: I think this can be made shorter for clarity?

```python
    @classmethod
    def load_from_checkpoint(*args, **kwargs):
        if kwargs.pop("baseline", "symnco") != "symnco":
            log.warning(
                "SymNCO only supports custom-symnco baseline. Setting to 'symnco'."
            )
        kwargs["baseline"] = "symnco"
        return super().load_from_checkpoint(*args, **kwargs)
```

n-g-z: I like this, but we should probably keep `checkpoint_path` in the params, as it is mandatory.

ngastzepeda: Another thing: we would lose type hints and IDE autocompletion, i.e., users would need to look at the parent class to know the parameter order and default values. So maybe we leave it as is, even though it's a bit longer?

fedebotu: Oh you're right! Then let's leave it as is~

fedebotu:
LGTM!

PR #264, opened 2025-04-09
Don't artificially limit the rendering axes for CVRP environment

### LINKED ISSUE #258: [BUG] plotting out of scale
## Describe the bug

CVRPTW generator scales both time and location coordinates with the same factor (max_time). This notably results in a bad rendering.

## To Reproduce

Initialize the CVRPTWGenerator with specific parameters, ensuring that scale=True and max_time is set to a particular value.​
Generate a set of instances using this generator.​
Examine the spatial distribution of the generated locations.

```python
from rl4co.envs.routing.cvrptw.env import CVRPTWEnv
env = CVRPTWEnv(generator_params={'num_loc':20,'scale':True})
env.render(td_init[0],actions=torch.tensor([0,1]))
```

## System info

rl4co : 0.5.2

## Additional context

The render function could also be updated (the axs limit currently assume you have scaled the problem)

## Reason and Possible fixes

Change this : 

if self.scale:
    durations = durations / self.max_time
    min_times = min_times / self.max_time
    max_times = max_times / self.max_time
    td["depot"] = td["depot"] / self.max_time
    td["locs"] = td["locs"] / self.max_time

To this : 
if self.scale:
    durations = durations / self.max_time
    min_times = min_times / self.max_time
    max_times = max_times / self.max_time
    td["depot"] = td["depot"] / self.max_loc
    td["locs"] = td["locs"] / self.max_loc

## Checklist

- [ ✅] I have checked that there is no similar issue in the repo (**required**)
- [ ✅] I have provided a minimal working example to reproduce the bug (**required**)

ngastzepeda:
## Description

Commented our the two lines of code that set the limits for the x and y axis, respectively.

## Motivation and Context

Assuming all locs would be scaled to [0,1] let to rendering problems for instances that were not in fact scaled to [0,1].

Closes #258 .

## Types of changes

## Checklist

### DIFF
--- a/rl4co/envs/common/base.py
+++ b/rl4co/envs/common/base.py
@@ -282,7 +282,9 @@ class RL4COEnvBase(EnvBase, metaclass=abc.ABCMeta):
 
     def render(self, *args, **kwargs):
         """Render the environment"""
-        raise NotImplementedError
+        raise NotImplementedError(
+            f"Render is not implemented for {self.name} environment. Please implement the `render` method in the subclass."
+        )
 
     @staticmethod
     def load_data(fpath, batch_size=[]):
--- a/rl4co/envs/eda/dpp/render.py
+++ b/rl4co/envs/eda/dpp/render.py
@@ -1,10 +1,7 @@
-import torch
-import numpy as np
 import matplotlib.pyplot as plt
+import numpy as np
+import torch
 
-from matplotlib import cm, colormaps
-
-from rl4co.utils.ops import gather_by_index
 from rl4co.utils.pylogger import get_pylogger
 
 log = get_pylogger(__name__)
@@ -15,7 +12,6 @@ def render(self, decaps, probe, action_mask, ax=None, legend=True):
     Plot a grid of 1x1 squares representing the environment.
     The keepout regions are the action_mask - decaps - probe
     """
-    import matplotlib.pyplot as plt
 
     settings = {
         0: {"color": "white", "label": "available"},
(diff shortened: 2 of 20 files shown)


fedebotu:
Sounds good!
I am also thinking about a more systematic change:
How about we never impose any artificial limit in the `render` function, but instead always return the `ax`?

For example:

```python
def render(td, ..., return_ax=False):
   # Main function
   # ... 
   if return_ax:
      return ax
```


In this way, users can directly modify the plot style (including setting the `x_lim` and `y_lim`) afterwards:


```python
ax = env.render(td, ..., return_ax=True):
ax.set_xlim(0, 1)
ax.set_ylim(0, 1)
```

WDYT?

ngastzepeda:
> Sounds good! I am also thinking about a more systematic change: How about we never impose any artificial limit in the `render` function, but instead return the `ax`?
> 
> For example:
> 
> ```python
> def render(td, ..., return_ax=False):
>    # Main function
>    # ... 
>    if return_ax:
>       return ax
> ```
> 
> In this way, users can directly modify the plot style (including setting the `x_lim` and `y_lim`) afterwards:
> 
> ```python
> ax = env.render(td, ..., return_ax=True):
> ax.set_xlim(0, 1)
> ax.set_ylim(0, 1)
> ```
> 
> WDYT?
> 
> cc @Estalisium, this should solve the plotting issue!

I have not tested this, but feel free to push to this branch @fedebotu :)

fedebotu:
Plotting bug from @Estalisium fixed, no more artificial restriction on range!
 
Also, now all `render` function calls return an `ax` by default!

This can be used to further modify the plot 

```python
ax = env.render(td[0], actions[0])
ax.set_xlim(-0.05, 1.05)
ax.set_ylim(-0.05, 1.05)
ax.set_xticks([])
ax.set_yticks([])
```

Since everything seems to be working well, I will proceed with merging the PR~

PR #282, opened 2026-08-19
fix: apply the gradient update in transductive search (EAS, ActiveSearch)

### LINKED ISSUE #281: [BUG] Transductive search (EAS, ActiveSearch) never applies its gradient update
## Describe the bug

`EAS` / `EASEmb` / `EASLay` and `ActiveSearch` compute a loss and backpropagate it, but never call `optimizer.step()`, so none of the parameters they are supposed to adapt is ever updated. Both methods currently reduce to plain (augmented, multi-start) sampling with incumbent re-injection.

There are two independent causes; either one alone breaks the update.

**1. `optimizer.step()` is missing.** In `EAS.training_step` (`rl4co/models/zoo/eas/search.py`) and `ActiveSearch.training_step` (`rl4co/models/zoo/active_search/search.py`):

```python
opt = self.optimizers()
opt.zero_grad()
self.manual_backward(loss)
# <- no opt.step()
```

The reference implementation ([`ahottung/EAS`](https://github.com/ahottung/EAS), `source/eas_emb.py`) does `optimizer.zero_grad()` / backward / `optimizer.step()`.

**2. `self.optimizers()` is not the optimizer holding the adapted parameters.** Both methods call `self.configure_optimizers(<subset of params>)` and discard the return value. `configure_optimizers` is a hook Lightning invokes once during `strategy.setup()`; calling it again by hand creates an optimizer that is never registered with the trainer. `self.optimizers()` therefore returns the optimizer built at setup time from `self.parameters()` — the whole policy, every parameter of which `EAS.on_train_batch_start` has just set to `requires_grad=False`.

So adding a bare `opt.step()` would step only frozen parameters whose `.grad` is `None`, and `opt.zero_grad()` never touches the adapted tensors, so their gradients accumulate across all search iterations instead of being reset.

## To Reproduce

`EASEmb` and `EASLay` adapt structurally different parameters, so with the same seed they must produce different results. They produce **bit-identical** results, which is only possible if neither of them adapts anything:

```python
import torch
from rl4co.envs import TSPEnv
from rl4co.models.zoo import EASEmb, EASLay
from rl4co.models.zoo.am import AttentionModelPolicy
from rl4co.utils.trainer import RL4COTrainer

results = {}
for cls in (EASEmb, EASLay):
    torch.manual_seed(0)
    env = TSPEnv(generator_params=dict(num_loc=20))
    policy = AttentionModelPolicy(env_name="tsp")
    model = cls(env, policy, env.dataset([4]), batch_size=2, max_iters=20, augment_size=8)
    RL4COTrainer(max_epochs=1, devices=1, logger=False, enable_checkpointing=False,
                 enable_progress_bar=False).fit(model)
    results[cls.__name__] = model.instance_rewards

print("EASEmb:", results["EASEmb"])
print("EASLay:", results["EASLay"])
print("identical:", torch.equal(results["EASEmb"], results["EASLay"]))
```

```bash
EASEmb: tensor([-5.4119, -6.5119, -5.8824, -6.1347], device='mps:0')
EASLay: tensor([-5.4119, -6.5119, -5.8824, -6.1347], device='mps:0')
identical: True
```

Instrumenting the tensor `EASEmb` adapts (`cached_embeds.logit_key`) confirms the mechanism directly — it has a non-zero gradient that is never zeroed, and it never moves:

```
adapted tensor: requires_grad=True  grad_norm=4.707e-01  max|delta|=0.000e+00
grad norm per iteration: 0.565 -> 0.844 -> 1.132 -> 1.299 -> 1.431 -> 1.635
self.optimizers() holds 82656 params, all with requires_grad=False
the optimizer built inside training_step holds 1280 params and is discarded
```

## Expected behavior

The parameters designated for adaptation (`cached_embeds.logit_key` for EAS-Emb, the `EASLayerNet` weights for EAS-Lay, all policy weights for `ActiveSearch`) should change during the search.

Note the rest of the EAS implementation matches the reference: all policy parameters are frozen; `EASEmb` adapts only `logit_key`, which is exactly POMO's `single_head_key` that official `eas_emb.py` optimizes; `EASLay` inserts the residual bottleneck on the MHA output before `project_out` with `W1`/`b1` Xavier- and `W2`/`b2` zero-initialised; and the loss is REINFORCE with a group-shared baseline over the sampled rollouts excluding the incumbent, plus `lambda * (-log p(incumbent))`. Only the optimizer step is missing.

## System info

- rl4co: reproduced on released `v0.6.0` and on `main` (`7a15396`); the files involved differ only cosmetically between the two
- torch 2.12.1, lightning 2.6.5, Python 3.13, macOS (also independent of accelerator/precision)

I have a fix ready and will open a PR referencing this issue.

*Disclaimer: both this report and the fix referenced above were co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 5). The reproduction script and every measurement quoted above were reviewed and run locally before submitting.*

MoritzOster:
## Description

Transductive search (`EAS`, `EASEmb`, `EASLay`, `ActiveSearch`) computes gradients but **never applies them**, so no parameter is ever updated. Both methods currently degenerate into plain (augmented, multi-start) sampling with incumbent re-injection.

There are two independent causes, and each one alone is sufficient to break the update:

**1. `optimizer.step()` is never called.** In `EAS.training_step` and `ActiveSearch.training_step` the manual-optimization block is

```python
opt = self.optimizers()
opt.zero_grad()
self.manual_backward(loss)
# no opt.step()
```

The reference implementation ([`ahottung/EAS`](https://github.com/ahottung/EAS), `source/eas_emb.py`) does `optimizer.zero_grad()` / backward / `optimizer.step()`. `git log -S "opt.step"` shows the call was never present in these two files.

**2. `self.optimizers()` does not return the optimizer that holds the adapted parameters.** Both methods call `self.configure_optimizers(<subset of params>)` and **discard the return value**. `configure_optimizers` is a hook that Lightning calls once during `strategy.setup()`; calling it again by hand builds an optimizer that is never registered with the trainer. So `self.optimizers()` returns the optimizer Lightning built at setup time — from `self.parameters()`, i.e. the *entire* policy, every parameter of which `EAS.on_train_batch_start` has just set to `requires_grad=False`.

Two consequences: adding a bare `opt.step()` would step only frozen parameters whose `.grad` is `None`, and `opt.zero_grad()` never touches the adapted tensors, so their gradients **accumulate across every search iteration**.

Measured on `EASEmb` / TSP-20 before the fix:

```
configure_optimizers calls (caller, #params): [('lightning-setup', 82656), ('inside-training_step', 1280)]
self.optimizers() is the EAS optimizer:       False
  Lightning's optimizer:  82656 params, requires_grad = {False}
  EAS's optimizer:         1280 params  (return value discarded)

adapted tensor: requires_grad=True  grad_norm=4.707e-01  max|delta|=0.000e+00
grad norm per iteration: 0.565 → 0.844 → 1.132 → 1.299 → 1.431 → 1.635   (never zeroed)
```

The shortest demonstration: on `main`, `EASEmb` and `EASLay` return **bit-identical** search trajectories, which is only possible if neither of them adapts anything (same seeds → same samples → same result). After the fix they differ from each other and from `main`:

```
                 tour length @ iter 0 / 10 / 20 / 39   (TSP50, 4 instances, seeded)
main    EASEmb    6.4282 / 6.3132 / 6.2735 / 6.2232
main    EASLay    6.4282 / 6.3132 / 6.2735 / 6.2232   <- identical to EASEmb
patch   EASEmb    6.4282 / 6.2971 / 6.2788 / 6.2357
patch   EASLay    6.4282 / 6.2984 / 6.2984 / 6.2236
```

## Fix

- **`TransductiveModel.setup_optimizer(parameters)`** (new, in `rl4co/models/common/transductive/base.py`): builds the optimizer over the adapted parameters and registers it via `self.trainer.strategy.optimizers = [optimizer]`, so it becomes the optimizer returned by `self.optimizers()`. Registering it (rather than using the raw `torch.optim.Optimizer`) matters: `RL4COTrainer` defaults to `precision="16-mixed"`, where `manual_backward` scales the loss and unscaling happens *inside* `LightningOptimizer.step()` → `precision_plugin.optimizer_step()`. Stepping a bare optimizer would apply gradients still scaled by the `GradScaler` factor.
- **`opt.step()`** added in both search loops.
- **`copy.deepcopy` for `original_policy_state`.** `self.policy.state_dict()` returns *references* to the live tensors, so `load_state_dict(self.original_policy_state)` in `on_train_batch_start` was a no-op (verified: identical `data_ptr()`). Harmless while nothing was being updated; once the update works this is a correctness bug — for `ActiveSearch`, which adapts all weights, instance *N+1* would otherwise start from instance *N*'s adapted policy instead of the pretrained one.
- **`self.policy.eval()` during search.** Search is a test-time procedure, but the policy was left in train mode, so batch-norm running statistics were updated by the search rollouts (12 buffers mutated in a 2-layer AM encoder) and — because of the aliasing above — the drift was never undone. Now 0 buffers are mutated.
- **Adapted embeddings are kept in the policy's parameter dtype.** Under mixed precision the encoder returns half-precision activations, so `nn.Parameter(cached_embeds.logit_key)` was a fp16 parameter. Once it is actually stepped, `GradScaler.unscale_` raises `ValueError: Attempting to unscale FP16 gradients` (and bf16 Adam updates of this magnitude would underflow). This is the standard AMP fp32-master-weights pattern, and matches what `EASLayerNet` already did implicitly. `EASLayerNet` is now created in the same dtype for consistency.

## Motivation and Context

`EAS` reproduces Hottung et al., ICLR 2022 (*Efficient Active Search for Combinatorial Optimization Problems*) and `ActiveSearch` reproduces Bello et al. (2016). Without a parameter update neither method does what it claims, and any numbers obtained from them are plain sampling results. The bug is present in released `v0.6.0` as well as on `main`.

Note the rest of the EAS implementation checks out against the reference: all policy parameters are frozen (`grouped_actor.requires_grad_(False)`); `EASEmb` adapts only `logit_key`, which is exactly POMO's `single_head_key` optimized by official `eas_emb.py`; `EASLay` inserts the residual bottleneck on the MHA output before `project_out` (= `multi_head_combine`) with `W1`/`b1` Xavier- and `W2`/`b2` zero-initialised; and the loss is REINFORCE with a group-shared baseline over the sampled rollouts *excluding* the incumbent, plus `λ · (−log p(incumbent))`. Only the optimizer step was missing.

close #<ISSUE_NUMBER>

- [x] I have raised an issue to propose this change ([required](https://github.com/ai4co/rl4co/issues) for new features and bug fixes)

## Types of changes

Not API-breaking, but it **does change the numerical output** of `EAS*` and `ActiveSearch` — they now actually search. Any previously recorded benchmark numbers for these methods should be regenerated.

## Testing

`tests/test_training.py::test_search_methods_update_adapted_parameters` (new) asserts that the parameters handed to `setup_optimizer` actually change during a search. It fails on the current code with `AssertionError: <Method> did not update any adapted parameter` for all three of `ActiveSearch`, `EASEmb`, `EASLay`, and passes with this patch. The existing `test_search_methods` smoke test still passes (including under the default `16-mixed` precision, which requires the dtype fix above).

To be explicit about what I could and could not verify: the evidence above establishes that the adapted parameters were never updated and now are. It does **not** establish a solution-quality improvement — my only checkpoint is a small AM I trained locally for a few epochs on TSP20, and on that checkpoint the 40-iteration TSP50 tour lengths above are a wash. That is expected (the default `lr`/`eas_lambda` are tuned for a POMO-class model with 200 iterations), but it does mean **the benchmarks need re-running with the official pretrained checkpoints** before trusting any EAS numbers from this repo.

## Known remaining issues (not addressed here, happy to open follow-ups)

Kept out of this PR to keep it reviewable; all pre-existing and independent of the gradient bug:

1. `EAS` only works with `batch_size >= 2` and `augment_size >= 2`. `unbatchify(..., (n_runs * batch_size, n_aug, group_s)).squeeze()` collapses singleton dimensions, then `reward.max(dim=2)` raises `IndexError: Dimension out of range`. Note `ActiveSearch` documents `batch_size=1`.
2. `num_parallel_runs > 1` raises — `best_solutions` is allocated as `[batch_size, ...]` but assigned `[n_runs * batch_size, ...]`. Relatedly, the `unbatchify` shape tuple assumes an `(n_aug, batch)` layout while `batchify(td_init, n_runs)` makes it `(n_runs, n_aug, batch)`, so it is only correct for `n_runs == 1`.
3. `EASLay` raises on environments with dynamic embeddings (e.g. SDVRP): the decoder batchifies the cache and keeps `td` flat at `[B*G]`, so `heads` is `[B*G, 1, E]` against `W1` of `[B, E, E]`. It works for TSP only via the static-embedding `unbatchify(td, num_starts)` branch.
4. The monkeypatches on `decoder.pointer.forward`, `decoder.pointer.eas_layer` and `decoder.forward_eas` are never reverted, so a policy is permanently modified by a search run. `forward_pointer_attn_eas_lay` is also a stale copy of `PointerAttention.forward`: it calls `self.project_out` instead of `self._project_out`, bypassing `PointerAttnMoE`'s hierarchical gating, and drops the `check_nan` assert.
5. `best_solutions[:, :len] = best_solutions_iter` does not clear the tail, so a shorter new incumbent keeps a stale suffix from the previous one (only matters for variable-length solutions, e.g. CVRP).
6. `EASEmb()` always logs a spurious "Setting `use_eas_embedding` to True..." warning (`kwargs.get("use_eas_embedding", False)` should default to `True`), and `EASLay`'s warning text is a copy of `EASEmb`'s. `verbose` is documented and accepted but never read.
7. Defaults are `lr=0.0041`, `eas_lambda=0.013`; the reference README uses `param_lr=0.0032` with `param_lambda=0.0058` (EAS-Emb) and `0.012` (EAS-Lay).

## Checklist

---

*Disclaimer: both this description and the code changes it describes were co-authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 5). The diagnosis, the code, and every measurement quoted above were reviewed and run locally before submitting.*

---

Closes #281

### DIFF
--- a/rl4co/models/common/transductive/base.py
+++ b/rl4co/models/common/transductive/base.py
@@ -2,7 +2,9 @@ import abc
 
 from typing import Any
 
+from lightning.pytorch.core.optimizer import LightningOptimizer
 from lightning.pytorch.utilities.types import STEP_OUTPUT
+from torch.optim import Optimizer
 from torch.utils.data import Dataset
 
 from rl4co.models.rl.common.base import RL4COLitModule
@@ -60,6 +62,28 @@ class TransductiveModel(RL4COLitModule, metaclass=abc.ABCMeta):
 
         # Setup loggers
         self.setup_loggers()
+
+    def setup_optimizer(self, parameters) -> LightningOptimizer:
+        """Instantiate an optimizer over `parameters` and register it with the Lightning
+        strategy, so that it is the optimizer returned by `self.optimizers()`.
+
+        Transductive methods only adapt a small subset of the parameters, which is not known
+        yet when Lightning calls the `configure_optimizers` hook at setup time. Registering
+        the optimizer here is what makes `self.optimizers().step()` update the adapted
+        parameters, and routes the step through the precision plugin so that mixed precision
+        (the `RL4COTrainer` default) unscales the gradients before stepping.
+
+        Args:
+            parameters: parameters to be adapted during the search
+        """
+        optimizer = self.configure_optimizers(parameters)
+        if not isinstance(optimizer, Optimizer):
+            raise ValueError(
+                "Transductive methods do not support learning rate schedulers; "
+                f"`configure_optimizers` must return a single optimizer, got {type(optimizer)}."
+            )
+        self.trainer.strategy.optimizers = [optimizer]
+        return self.optimizers()
 
     def on_train_batch_start(self, batch: Any, batch_idx: int):
         """Called before training (i.e. search) for a new batch begins.
--- a/rl4co/models/zoo/active_search/search.py
+++ b/rl4co/models/zoo/active_search/search.py
@@ -1,3 +1,4 @@
+import copy
 import time
 
 from functools import partial
@@ -86,7 +87,7 @@ class ActiveSearch(TransductiveModel):
         )
 
         # Store original policy state dict
-        self.original_policy_state = self.policy.state_dict()
+        self.original_policy_state = copy.deepcopy(self.policy.state_dict())
 
         # Get dataset size and problem size
         dataset_size = len(self.dataset)
@@ -100,7 +101,12 @@ class ActiveSearch(TransductiveModel):
         We re-load the original policy state dict and configure the optimizer.
         """
         self.policy.load_state_dict(self.original_policy_state)
-        self.configure_optimizers(self.policy.parameters())
+
+        # Search happens at test time: keep the policy in eval mode so that e.g. batch
+        # normalization uses its running statistics instead of updating them
+        self.policy.eval()
+
+        self.setup_optimizer(self.policy.parameters())
 
     def training_step(self, batch, batch_idx):
         """Main search loop. We use the training step to effectively adapt to a `batch` of instances."""
@@ -155,6 +161,7 @@ class ActiveSearch(TransductiveModel):
             opt = self.optimizers()
             opt.zero_grad()
             self.manual_backward(loss)
+            opt.step()
 
             self.log_dict(
                 {
--- a/rl4co/models/zoo/eas/search.py
+++ b/rl4co/models/zoo/eas/search.py
@@ -1,3 +1,4 @@
+import copy
 import time
 
 from functools import partial
@@ -114,7 +115,7 @@ class EAS(TransductiveModel):
         )
 
         # Store original policy state dict
-        self.original_policy_state = self.policy.state_dict()
+        self.original_policy_state = copy.deepcopy(self.policy.state_dict())
 
         # Get dataset size and problem size
         len(self.dataset)
@@ -129,6 +130,10 @@ class EAS(TransductiveModel):
         We do the rest in the training step.
         """
         self.policy.load_state_dict(self.original_policy_state)
+
+        # Search happens at test time: keep the policy in eval mode so that e.g. batch
+        # normalization uses its running statistics instead of updating them
+        self.policy.eval()
 
         # Set all policy parameters to not require gradients
         for param in self.policy.parameters():
@@ -160,11 +165,19 @@ class EAS(TransductiveModel):
         embeddings, _ = encoder(td_init)
         cached_embeds = decoder._precompute_cache(embeddings)
 
+        # Keep the adapted parameters in the same dtype as the policy parameters: under mixed
+        # precision the encoder returns half precision activations, which must not be optimized
+        # directly (the gradient scaler refuses to unscale fp16 gradients, and half precision
+        # optimizer updates underflow). This is the usual AMP pattern of fp32 master weights.
+        param_dtype = next(self.policy.parameters()).dtype
+
         # Collect optimizer parameters
         opt_params = []
         if self.hparams.use_eas_layer:
             # EASLay: replace forward of logit attention computation. EASLayer
-            eas_layer = EASLayerNet(num_instances, decoder.embed_dim).to(batch.device)
+            eas_layer = EASLayerNet(num_instances, decoder.embed_dim).to(
+                device=batch.device, dtype=param_dtype
+            )
             decoder.pointer.eas_layer = partial(eas_layer, decoder.pointer)
             decoder.pointer.forward = partial(forward_pointer_attn_eas_lay, decoder.pointer)
             for param in eas_layer.parameters():
@@ -173,7 +186,11 @@ class EAS(TransductiveModel):
             # EASEmb: set gradient of emb_key to True
             # for all the keys, wrap the embedding in a nn.Parameter
             for key in self.hparams.eas_emb_cache_keys:
-                setattr(cached_embeds, key, torch.nn.Parameter(getattr(cached_embeds, key)))
+                setattr(
+                    cached_embeds,
+                    key,
+                    torch.nn.Parameter(getattr(cached_embeds, key).to(param_dtype)),
+                )
                 opt_params.append(getattr(cached_embeds, key))
         decoder.forward_eas = partial(forward_eas, decoder)
 
@@ -185,7 +202,7 @@ class EAS(TransductiveModel):
         for attr in ["temperature", "tanh_clipping", "mask_logits"]:
             set_attr_if_exists(attr)
 
-        self.configure_optimizers(opt_params)
+        opt = self.setup_optimizer(opt_params)
 
         # Solution and reward buffer
         max_reward = torch.full((batch_size,), -float("inf"), device=batch.device)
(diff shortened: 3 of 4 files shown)


fedebotu:
Thank you @MoritzOster ! Reproduced on main: EAS and ActiveSearch never updated anything, so they were effectively just sampling. with the fix, EASLay finds ~10% shorter tours on TSP200 than without the update. We trimmed it a bit (dropped the policy.eval() change, which is a separate behavior change, plus the scheduler check and the EASLayerNet dtype cast, and shortened the comments) and re-ran the transductive notebook. Now it seems to be good to go -- will merge soon!

PR #284, opened 2026-09-09
fix: correct SymNCO invariance loss sign

### LINKED ISSUE #283: [BUG] SymNCO invariance loss minimizes cosine similarity
## Describe the bug

SymNCO's `invariance_loss` returns positive cosine similarity, which is added to the minimized training loss with a positive default weight (`alpha=0.2`). This favors dissimilar projected representations.

## To Reproduce

```python
import torch
from rl4co.models.zoo.symnco.losses import invariance_loss

aligned = torch.tensor([[[1., 0.]], [[1., 0.]]])
opposite = torch.tensor([[[1., 0.]], [[-1., 0.]]])
print(invariance_loss(aligned, 2))   # +1
print(invariance_loss(opposite, 2))  # -1
```

Run against upstream commit `7a1539678b6a81cc68b4069440aa5e98f6931f6c` before the fix. Aligned representations should have a lower loss.

## System info

RL4CO 0.6.0 (source checkout), Python 3.12.13, PyTorch 2.6.0+cpu, TorchRL 0.6.0, TensorDict 0.6.0, Linux.

## Reason and Possible fixes

Return `-similarity.mean()` in `rl4co/models/zoo/symnco/losses.py`.

## Additional context

A maintainer confirmed the sign issue and suggested submitting a fix.

## Checklist

- [x] I have checked that there is no similar issue in the repo.
- [x] I have provided a minimal working example to reproduce the bug.

keep9oing:
## Description

Negate SymNCO's invariance loss so minimizing it increases cosine similarity between projected representations.

## Motivation and Context

The current positive similarity term favors dissimilar representations. Regression tests check loss ordering and the gradient direction for two and four augmentations.

- [x] I have raised an issue to propose this change.

## Types of changes

- [x] Bug fix

## Checklist

- [x] I have updated the tests accordingly.

## Validation

- `pytest -q tests`: 133 passed (CPU).
- Pre-commit checks passed for both changed files.
- All four regression cases fail before the fix and pass after it.

Closes #283

### DIFF
--- a/rl4co/models/zoo/symnco/losses.py
+++ b/rl4co/models/zoo/symnco/losses.py
@@ -29,9 +29,9 @@ def solution_symmetricity_loss(reward, log_likelihood, dim=-1):
 
 
 def invariance_loss(proj_embed, num_augment):
-    """Loss for invariant representation on projected nodes
+    """Negative cosine similarity loss for invariant projected node representations.
     Corresponds to `L_inv` in the SymNCO paper
     """
     pe = rearrange(proj_embed, "(b a) ... -> b a ...", a=num_augment)
     similarity = sum([cosine_similarity(pe[:, 0], pe[:, i], dim=-1) for i in range(1, num_augment)])
-    return similarity.mean()
+    return -similarity.mean()
--- a/rl4co/models/zoo/symnco/losses.py
+++ b/rl4co/models/zoo/symnco/losses.py
@@ -1,5 +1,6 @@
-from einops import rearrange
 from torch.nn.functional import cosine_similarity
+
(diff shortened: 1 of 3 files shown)


fedebotu:
Thank you @keep9oing , good catch!

There was a second pre-existing bug in the same line that had never been caught: `batchify` lays the batch out as (a b), not (b a), so the loss was comparing embeddings from different instances. I pushed a fix that uses `unbatchify` (matching the official `reshape(N_aug, -1, ...))`. It went unnoticed because this term is quite small either way and never really influenced the losses much :eyes: 

Merging now

